PR #32, opened 2022-02-28
Fewer string allocations

Kijewski:
The longest time zone name has 25 characters. Using `String::with_capacity()` it is possible have fewer re-allocations, and possible systems calls.

### DIFF
--- a/src/tz_linux.rs
+++ b/src/tz_linux.rs
@@ -1,11 +1,7 @@
 pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
     // see https://stackoverflow.com/a/12523283
-    use std::io::Read;
-
-    let fname = "/etc/timezone";
-    let mut f = std::fs::File::open(&fname)?;
-    let mut contents = String::with_capacity(32);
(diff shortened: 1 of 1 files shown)


astraw:
What do you think about https://github.com/strawlab/iana-time-zone/commit/d830c3d554cdff8c5e151d83f1d827c0fe94b510 instead? It defers the string allocation to `std::fs::read_to_string` which looks like it uses the file metadata to reserve exactly what is needed. But perhaps the `.trim().into()` then causes reallocation?

Kijewski:
`s.trim().into()` would allocate another string (not reallocate). Using read_to_string() is good, but I'd still use `s.truncate(s.trim_end().len())`, unless there can be spaces at the front, tot. But I guess that's invalid?

astraw:
Great. 

I released this in 0.1.30. Thanks.

astraw:
(And, to answer your question, yes, I am assuming there is no whitespace at the start of the string.)

Kijewski:
> I released this in 0.1.30. Thanks.

Thank you for your fast replies! :)

PR #35, opened 2022-04-12
Use winapi crate instead of windows crate

Kijewski:
This reduces the minimum supported rust version by quite a lot.


astraw:
Yes, I was a little surprised that `windows` 0.35 updated their MSRV to 1.59 recently, which is what caused me to take the opportunity to update the MSRV in iana-time-zone also to 1.59. (Although for mac and linux, this wasn't actually necessary.)

But with this PR, I am concerned that increases dramatically the amount of code to be maintained here. Also, I assumed due to Microsoft's own support that [`windows`](https://crates.io/crates/windows) was "the future" but I don't actually follow this very closely. I guess `winapi` and `windows` crates are two ways of doing the same thing, one widely used and independently maintained (`winapi`) and the other less widely used for now and maintained by microsoft (`windows`)?

At the time of bumping the MSRV here, I checked into the issue "does updating MSRV imply the semver should be bumped to indicate a breaking change?" to which I found the general consensus to be "no" because a) people can painlessly update, b) otherwise people have `Cargo.lock` and c) otherwise the ecosystem gets split. The other general point I looked into was an answer to the question "what reasons to people have for sticking with an old rust version that cannot be addressed fairly easily with a `Cargo.lock` file?" to which the consensus I found to be "few good reasons". With that research done, I updated my rust compiler in CI and then bumped the MSRV in the libraries and moved on...

Due to those reasons, I am hesitant about the switch to `winapi` but the other changes here are certainly interesting. What do you think?

Kijewski:
It's difficult. I for one am not even sure what's stopping people from simply using the most current stable rust version, but I guess there's a bunch of Centos users out there. :)

The windows crate (probably) is the future, but I think it's still in its infancy with many breaking changes between versions. (But don't take my word for it, because I'm neither a windows user nor do a program for windows normally.) It's too bad that the winapi crate is not supported by a team, esp. since its used by fundamental crates like tokio.

My code is ugly, but I don't think that there is the need to ever touch it again. I added few comments, but I don't think that it's easy to read or understand. I would totally understand if you choose not to adopt my PR.

Kijewski:
I included the Cargo.toml so cargo audit works.

astraw:
The argument that tokio uses `winapi` is enough for me. And indeed the constant breaking changes in `windows` (although it has slowed lately) is a bit tiring. There are also many other nice things here and in the absence of a reason to really *need* a recent rust compiler, I agree it is just nicer to support older versions.

niyaznigmatullin:
It wasn't mentioned here I hope you are aware, but there is `windows-sys` crate with MSRV `1.49` and is lightweight version of `windows` crate, which is maintained unlike `winapi`.

Kijewski:
When you look at the [history of their Cargo.lock](https://github.com/microsoft/windows-rs/commits/master/crates/libs/sys/Cargo.toml), you'll see that they randomly bump their msrv every other release. It's still quite an old rust version they need, but I don't think their release pattern makes the `windows-sys` crate any more useful than the `windows` crate itself.

niyaznigmatullin:
You are right about that. To defend the crate, the maintainers seem to have some thoughts on this matter: https://github.com/microsoft/windows-rs/pull/1987

 _(more of a joke: considering only this point, `winapi` is certainly better, because they will never bump their MSRV, since there will be no release in the future :))_

PR #37, opened 2022-04-14
Remove need for reallocations for MacOS implementation

Kijewski:


.github/workflows/rust.yml
```diff
@@ -37,6 +37,7 @@ jobs:
             ~/.cargo/git/db/
             target/
       - run: cargo test --workspace --all-targets
+      - run: cargo run --example get_timezone
```
Kijewski, line 40: I don't know how to change the timezone of the runner, so don't know if there is a better option than to at the result whether it looks sane.

src/tz_macos.rs
```diff
@@ -4,13 +4,17 @@ pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
     // Get string like ""Europe/Berlin (GMT+2) offset 7200 (Daylight)""
     let mut str1 = format!("{:?}", tz);
 
-    // strip leading double quotes
-    while str1.starts_with('"') {
-        str1 = str1[1..].to_string();
-    }
+    // We split the string first, so removing the leading quotes has to move fewer characters.
+    str1.truncate(match str1.split_ascii_whitespace().next() {
+        Some(s) => s.len(),
```
Kijewski, line 9: I put the truncation outside of the match. Inside the Some() arm, the string is still borrowed. Cf. https://github.com/rust-lang/rust/issues/59159

astraw:
I had a look at this myself and discovered the [`core_foundation::timezone::CFTimeZone::name()`](https://docs.rs/core-foundation/latest/core_foundation/timezone/struct.CFTimeZone.html#method.name) method which solves the problem more elegantly I think. So this is superseded by 9acb23308ea4b50b2bf7fed27f63ded203098764.

Thanks for prompting me to look into this.

Recent changes released as 0.1.33.

Kijewski:
Nice, that much prettier!

PR #38, opened 2022-06-29
Implement for wasm

Kijewski:
<code>[Intl.DateTimeFormat()]</code> makes the timezone of the browser or node instance queryable, and [is available] in all browsers that support WASM.

[Intl.DateTimeFormat()]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat
[is available]: https://caniuse.com/mdn-javascript_builtins_intl_datetimeformat_resolvedoptions

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -56,7 +56,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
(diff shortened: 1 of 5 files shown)


astraw:
Nice. thanks. Released as 0.1.34.

PR #39, opened 2022-06-29
Implement for {Free,Net,Open}BSD and Dragonfly

Kijewski:
Tested successfully on FreeBSD and NetBSD. Dragonfly should work the same as FreeBSD, and OpenBSD as NetBSD.

There are no Github action runners for either of these four BSDs, so the examples get cross compiled, but not executed.


Kijewski:
As far as I can tell now only Android, Fuchsia, Ios, Illumos, and Redox are missing to complete the [tier 2 platform](https://doc.rust-lang.org/nightly/rustc/platform-support.html) list.

PR #40, opened 2022-07-21
Fail to compile for WASI

Kijewski:
PR #38 did not properly separate WebAssembly for the use in a
browser/nodejs, or standalone run in e.g. wasmtime. This PR make the
compilation fail if the target-os is WASI.

Also I deduplicated the imports by using `#[cfg_attr(…, path = …)]`
instead of importing and aliassing in two steps.


astraw:
Nice. Thanks, published as 0.1.36.

PR #41, opened 2022-07-22
Support iOS

Kijewski:
Cf. https://github.com/baoyachi/shadow-rs/pull/107#issuecomment-1192489565

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -109,6 +109,39 @@ jobs:
       - run: which cross || cargo install cross
       - run: cross build --target ${{ matrix.target }} --examples
 
+  build-ios-cross:
+    strategy:
+      matrix:
+        target: [aarch64-apple-ios-sim, aarch64-apple-ios, x86_64-apple-ios]
+    runs-on: macos-12
+    name: "Build for ${{ matrix.target }}"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install Rust
+        id: actions-rs
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+          target: ${{ matrix.target }}
+      - name: Cache
+        uses: actions/cache@v3
+        with:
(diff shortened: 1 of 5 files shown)


Kijewski:
I haven't tested this PR, not having an iOS device. @appaquet, can you please have a look if that fixes the problem for you?

appaquet:
Your branch now compiles when targeting iOS. Could be a good idea to add MacOS and iOS checks in the GitHub actions flow! I didn't test the functionally though, but since it's using core foundation for both iOS and MacOS, it should be fine in both sets of targets.

Kijewski:
Would you be able to test if it actually works, though?

No iOS target is supported by cross, so I don't know how I could add a test. If you do, please tell.

appaquet:
I'll try to test this later in my project. 

Yeah, you can't really cross compile to iOS with cross since it needs the toolchains to do so, which are available on MacOS with Xcode, but you can't redistribute them. You can actually compile for iOS on GitHub actions though (they have macos runners), but you won't be able to run your tests since you need to wrap it into an application to run it in a simulator.

Kijewski:
Ah, thanks for the information. It does cross compile fine on a macOS-12 runner.

astraw:
The tests run on macOS, including automatically by Github Actions. The code is the same for iOS and macOS, so I guess it works on iOS as long as it compiles but I haven't tested it.

This PR adds a new variant to the `GetTimezoneError` enum, which will be a breaking change. Is it worth it? I'm not convinced. Also I think I prefer the compile-time error rather than the run-time error when my platform is not supported. Perhaps a middle ground is that the relevant function is simply not defined, which will result in a compile time error but only if the function is called.

Kijewski:
You're right. I removed the new variant.

astraw:
> How can you do that?

Ahh, I mis-remembered the old behavior and just played with it now. It doesn't behave like I thought, so please ignore this part of my comment.

Kijewski:
I restored the old behavior, so it still compile_errors.

appaquet:
@Kijewski  As expected, just tested in an iOS sim on `aarch64-apple-ios-sim` and it works.

Kijewski:
Thank you for testing it, and reporting the issue!

astraw:
published in release 0.1.37 . Thanks all.

PR #42, opened 2022-07-22
Implement `std::err::source()`, format IoError

Kijewski:


baoyachi:
Cargo.lock may be add git .gitignore . Not must commmit

astraw:
published in release 0.1.37. Also removed `Cargo.lock`. Thanks all.

PR #43, opened 2022-07-26
WIP: implement working in a no-std context

Kijewski:
Changes to API not use dynamic allocations, so it can work with `#![no_std]`. The API is *mostly* the same, but a stack allocating, limited length string type is used to return the timezone. Raises the msrv to 1.48.


astraw:
Where would the timezone information come from on a no-std target? In other words, what's the use case for this?

Regarding the implementation here, at a high level it seems like the `crate::timezone::Timezone` type has substantial overlap with the &str type. So, I wonder if we would be better served by a function which would modify a (heap or stack) allocated `&str` like the following. This is just an idea for discussion.

```rust
/// get the timezone and place it into `dest`, returning the length
pub fn get_timezone_into(mut dest: &mut str) -> Result<usize, crate::GetTimezoneError> {
    platform::get_timezone_into(dest)
}
```

Then, on platforms with `std`, we could keep the existing signature:

```rust
#[cfg(feature="alloc")]
pub fn get_timezone() -> Result<String, i8> {
    let mut result = String::from("                         "); // TODO: ensure long enough
    let len = get_timezone_into(result.as_mut_str())?;
    result.truncate(len);
    Ok(result)
}
```

This would seem to prevent the need for a breaking change.

Kijewski:
It does not have to be, that you cannot use std, but maybe you don't want to. Maybe to want to build a tiny program, reduce the build time.

Also on Linux, with the PR the stress-test runs in (wall-clock time, best of three runs) 0.838s. In v0.1.37 it runs in 0.931s. Most likely because I use rustix, and a stack allocated, presized buffer instead of `std::fs::read_to_string()`, which uses the heap. Not using `read_to_string()`, but a buffer + string conversion might see the same speed up.

You are right, the old interface can (and should be) kept. I changed that in the last commit. `Timezone` is not much more than a pre-allocated buffer, but the user does not have to know what the appropriate length is. Most of it simply implements standard interfaces, so it can be used like an `&str`.

src/tz_windows.rs
```diff
@@ -27,43 +32,35 @@ const WINDOWS_GLOBALIZATION_CALENDAR: &[wchar_t] = &wstring!(
     0
 );
 
-static INITALIZED: sync::Once = sync::Once::new();
-static mut FACTORY: Result<Unknown<IActivationFactory>, HRESULT> = Err(0);
+static mut FACTORY: OnceCell<Result<Unknown<IActivationFactory>, HRESULT>> = OnceCell::new();
```
lopopolo, line 35: this doesn't need to be a `static mut` anymore I believe since `OnceCell` gives interior mutability.

pinkforest:
This is probably embedded use-case and wasm32-unknown-unknown ? wasi provides system interface but it's via std

Probably best to have a strategy around supported target triplets and then design featuresets around those -

You could use [heapless](https://crates.io/crates/heapless) if you don't want to use alloc crate which comes via no_std crates

I mean often the baremetal consists of option to either go heapless or with alloc that then provides the heap allocated types

Real problem will be the system interface to derive information from -

You could also feature gate these out - e.g. often serialisizing crates provide "alloc" and default-features = false is essentially no_std and std feature triggers std use (default feature)

You can look at [postcard](https://crates.io/crates/postcard) crate for how the `heapless` and `alloc` works together with `serde` to featuregate whether someone wants to use either model.

src/tz_windows.rs
pinkforest: core (no_std) has OnceCell now.. but it's still not stable.. maybe feature gate it ?
https://doc.rust-lang.org/nightly/core/cell/struct.OnceCell.html

Kijewski:
Thank you for reviewing this. I'm not currently intending so work on this PR in the foreseeable time, I just have too many open projects that have much higher priority to me.

I think exactly every file in this project was touched since this PR was first started, so I guess I would have to start from scratch anyway.

Kijewski:
I'm current not working on this, and since I last touched the PR almost every line was changed. If I or anyone else ever finds the time to work on this, I think opening a new PR would be better.

PR #45, opened 2022-07-28
Implement for Android

licy183:
Tested on x86_64-linux-android/aarch64-linux-android/arm-linux-androideabi.

### DIFF
--- a/src/tz_android.rs
+++ b/src/tz_android.rs
@@ -1,0 +1,12 @@
+use android_system_properties::AndroidSystemProperties;
+
+// From https://android.googlesource.com/platform/ndk/+/android-4.2.2_r1.2/docs/system/libc/OVERVIEW.html
+// The system property named 'persist.sys.timezone' contains the name of the current timezone.
+
+const TIMEZONE_PROP_KEY: &str = "persist.sys.timezone";
+
+pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
+    AndroidSystemProperties::new()
+        .get(TIMEZONE_PROP_KEY)
+        .ok_or(crate::GetTimezoneError::OsError)
+}
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -83,6 +83,10 @@ jobs:
           - x86_64-unknown-freebsd
           - x86_64-unknown-illumos
           - x86_64-unknown-netbsd
+          - x86_64-linux-android
+          - i686-linux-android
+          - arm-linux-androideabi
+          - aarch64-linux-android
     runs-on: ubuntu-latest
     name: "Build for ${{ matrix.target }}"
     steps:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,9 @@ categories = ["date-and-time", "internationalization", "os"]
 readme = "README.md"
 edition = "2018"
 
+[target.'cfg(target_os = "android")'.dependencies]
+android_system_properties = "0.1.4"
+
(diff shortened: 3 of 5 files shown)


src/tz_android.rs
```diff
@@ -0,0 +1,24 @@
...
+// From https://android.googlesource.com/platform/ndk/+/android-4.2.2_r1.2/docs/system/libc/OVERVIEW.html
+// The name of the current timezone is taken from the TZ environment variable,
+// if defined. Otherwise, the system property named 'persist.sys.timezone' is
+// checked instead.
+
+const TIMEZONE_ENV_KEY: &str = "TZ";
+const TIMEZONE_PROP_KEY: &str = "persist.sys.timezone";
+
+pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
+    return match env::var(TIMEZONE_ENV_KEY) {
```
Kijewski, line 13: Actually, we don't check the environment variables for any other OS, and unless Android is **very** special in this regard, I don't think we should do it in here, either. Cf: https://github.com/time-rs/time/security/advisories/GHSA-wcg3-cvx6-7396

src/tz_android.rs
```diff
@@ -0,0 +1,24 @@
...
+        Ok(val) => Ok(val),
+        _ => {
+            let properties = AndroidSystemProperties::new();
+            if let Some(value) = properties.get(TIMEZONE_PROP_KEY) {
+                Ok(value)
+            } else {
+                Err(crate::GetTimezoneError::OsError)
+            }
```
Kijewski, line 21: This can be simplified to

```rust
properties.get(TIMEZONE_PROP_KEY).ok_or(crate::GetTimezoneError::OsError)
```

Kijewski (changes requested):
Thank you for the PR! I only have two minor comments.

src/tz_android.rs
licy183: I'm not sure about this. `TZ` environment is not set by default in Android, so let's assume it that `TZ` is always not set. I will remove this later.

src/tz_android.rs
licy183: Ok, I'll fix this.

src/tz_android.rs
```diff
@@ -0,0 +1,14 @@
+use std::env;
+use android_system_properties::AndroidSystemProperties;
+
+// From https://android.googlesource.com/platform/ndk/+/android-4.2.2_r1.2/docs/system/libc/OVERVIEW.html
+// The name of the current timezone is taken from the TZ environment variable,
+// if defined. Otherwise, the system property named 'persist.sys.timezone' is
+// checked instead.
```
Kijewski, line 7: Could you please update the comment, too?

src/tz_android.rs
```diff
@@ -0,0 +1,14 @@
...
+
+const TIMEZONE_PROP_KEY: &str = "persist.sys.timezone";
+
+pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
+    let properties = AndroidSystemProperties::new();
+    return properties.get(TIMEZONE_PROP_KEY).ok_or(crate::GetTimezoneError::OsError);
```
Kijewski, line 13: In Rust, the outcome of the last statement is also the result of the function. So if the last statement it `return something;`, you can simplify it to `something`:

```rust
    AndroidSystemProperties::new()
        .get(TIMEZONE_PROP_KEY)
        .ok_or(crate::GetTimezoneError::OsError)
}
```

Kijewski (changes requested):

astraw:
published in 0.1.40

licy183:
Thanks!

PR #48, opened 2022-08-10
Fix implementation for Redhat based distros

### LINKED ISSUE #47: Linux implementation does not work on CentOS 7
According to https://unix.stackexchange.com/a/451714/7851 systemd's `timedatectl` does not always update `/etc/timezone`, and it uses `/etc/localtime` instead. So most likely `/etc/localtime` should overrule `/etc/timezone` on systemd based distros, which is most of them.

Maybe the simplest solution would be to merge [tz_netbsd.rs](https://github.com/strawlab/iana-time-zone/blob/main/src/tz_netbsd.rs) and [tz_linux.rs](https://github.com/strawlab/iana-time-zone/blob/main/src/tz_linux.rs), and first try the current netbsd implemention, then the current linux implementation (which seems to be Debian and its offspring specific).

Related issue: https://github.com/Kijewski/tzdb/issues/101

Kijewski:
According to <https://unix.stackexchange.com/a/451714/7851> systemd's
`timedatectl` does not always update `/etc/timezone`, and it uses
`/etc/localtime` instead. So most likely `/etc/localtime` should
overrule `/etc/timezone` on systemd based distros, which is most of
them.

This PR makes the Linux implementation first try to parse
`/etc/localtime`, and if that fails `/etc/timezone`.

This actually makes the implementation a lot faster. The old
implementation that only checked `/etc/timezone` runs in

```text
$ time target/release/examples/stress-test

real	0m0,910s
user	0m0,630s
sys	0m7,909s
```

and the new implementation runs in

```text
$ time target/release/examples/stress-test

real	0m0,197s
user	0m0,345s
sys	0m1,453s
```

on a f2fs formatted root partition. The speed advantage might be less or
not be present at all with other file system.

Closes  #47.

### DIFF
--- a/src/tz_linux.rs
+++ b/src/tz_linux.rs
@@ -1,7 +1,31 @@
+use std::fs::{read_link, read_to_string};
+
 pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
+    etc_localtime().or_else(|_| etc_timezone())
+}
+
+fn etc_timezone() -> Result<String, crate::GetTimezoneError> {
     // see https://stackoverflow.com/a/12523283
-    let mut contents = std::fs::read_to_string("/etc/timezone")?;
+    let mut contents = read_to_string("/etc/timezone")?;
     // Trim to the correct length without allocating.
     contents.truncate(contents.trim_end().len());
     Ok(contents)
+}
+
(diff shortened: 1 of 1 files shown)


astraw:
Thanks. Released in 0.1.42.

PR #55, opened 2022-08-18
Implement for Solaris

Kijewski:
Untested, because `x86_64-sun-solaris` is an alias for
`x86_64-unknown-illumos` by now, and there is no docker image for
`x86_64-pc-solaris`.


astraw:
LGTM. Happy if you merge and publish this.

PR #56, opened 2022-08-30
Update to android_system_properties v0.1.5

Kijewski:
The new release makes `AndroidSystemProperties` send + sync, so the
initialization can be cached. This makes the stress test execute in

```text
0m00.61s real     0m04.00s user     0m00.02s system
```

Before it was

```text
1m37.25s real     1m06.81s user     0m57.95s system
```

I.e. it's now 9786 % faster. :)

### DIFF
--- a/src/tz_android.rs
+++ b/src/tz_android.rs
@@ -1,12 +1,16 @@
+use std::ffi::CStr;
+
 use android_system_properties::AndroidSystemProperties;
+use once_cell::sync::OnceCell;
 
 // From https://android.googlesource.com/platform/ndk/+/android-4.2.2_r1.2/docs/system/libc/OVERVIEW.html
 // The system property named 'persist.sys.timezone' contains the name of the current timezone.
 
-const TIMEZONE_PROP_KEY: &str = "persist.sys.timezone";
+static PROPERTIES: OnceCell<AndroidSystemProperties> = OnceCell::new();
 
 pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
-    AndroidSystemProperties::new()
-        .get(TIMEZONE_PROP_KEY)
+    PROPERTIES
+        .get_or_init(AndroidSystemProperties::new)
+        .get_from_cstr(unsafe { CStr::from_bytes_with_nul_unchecked(b"persist.sys.timezone\0") })
         .ok_or(crate::GetTimezoneError::OsError)
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.47] - 2022-08-30
+### Changed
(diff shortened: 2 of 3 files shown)


astraw:
I don't think we should cache results from the OS, as this would prevent getting an updated time zone in a long running process. With mobile OSes like Android in particular I guess this is an important issue. If better performance is needed, the user can cache themselves.

Kijewski:
This does not cache the result, only the "getter". We do the same on already in the windows implementation. The getter is still called in every invocation.

astraw:
Thanks for the clarification. Yes, now that my coffee has kicked in, I see that, too. This looks good, so I merged it and published as 0.1.47.

src/tz_android.rs
```diff
@@ -1,12 +1,16 @@
+use std::ffi::CStr;
+
 use android_system_properties::AndroidSystemProperties;
+use once_cell::sync::OnceCell;
 
 // From https://android.googlesource.com/platform/ndk/+/android-4.2.2_r1.2/docs/system/libc/OVERVIEW.html
 // The system property named 'persist.sys.timezone' contains the name of the current timezone.
 
-const TIMEZONE_PROP_KEY: &str = "persist.sys.timezone";
+static PROPERTIES: OnceCell<AndroidSystemProperties> = OnceCell::new();
```
lopopolo, line 9: Can we use a `once_cell::sync::Lazy` here? I find that a bit easier to read at call sites.

src/tz_android.rs
```diff
@@ -1,12 +1,16 @@
...
 
 pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
-    AndroidSystemProperties::new()
-        .get(TIMEZONE_PROP_KEY)
+    PROPERTIES
+        .get_or_init(AndroidSystemProperties::new)
+        .get_from_cstr(unsafe { CStr::from_bytes_with_nul_unchecked(b"persist.sys.timezone\0") })
```
lopopolo, line 14: Did the API change in the android crate that requires the unsafe here?

Kijewski: I introduced the new method in https://github.com/nical/android_system_properties/pull/9. `AndroidSystemProperties::get()` still exists, and the new `AndroidSystemProperties::get_from_cstr()` is safe, too. The new function does not alloc a CString at runtime to get a properly null terminated `&CStr`.

[`CStr::from_bytes_with_nul_unchecked()`](https://doc.rust-lang.org/stable/core/ffi/struct.CStr.html#method.from_bytes_with_nul_unchecked) is unsafe, because you have to ensure that it ends with a `\0`, and there are no `\0` inside the string except for the terminator. I could have used the save method [`CStr::from_bytes_with_nul()`](https://doc.rust-lang.org/stable/core/ffi/struct.CStr.html#method.from_bytes_with_nul), but well. I trust my eyes that I see a `\0` at the end of the string. :)

src/tz_android.rs
Kijewski: Actually, to me that's the opposite, at least for one liner initializations. To me ↓ almost reads like a sentence:

```rust
PROPERTIES.get_or_init(AndroidSystemProperties::new)
```

lopopolo: Sounds good

PR #57, opened 2022-09-05
Support AIX operating system

ecnelises:
This change makes this crate support detecting timezone on AIX. Since AIX support hasn't been merged yet, this is in draft status.

### DIFF
--- a/src/tz_aix.rs
+++ b/src/tz_aix.rs
@@ -1,0 +1,26 @@
+use std::fs::OpenOptions;
+use std::io::{BufRead, BufReader};
+use std::env;
+
+pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
+    env::var("TZ").map_err(|_| crate::GetTimezoneError::OsError)
+}
+
+fn read_environment() -> Result<String, crate::GetTimezoneError> {
+    // https://www.ibm.com/docs/en/aix/7.2?topic=files-environment-file
+
+    let file = OpenOptions::new().read(true).open("/etc/environment")?;
+    let mut reader = BufReader::new(file);
+    let mut line = String::with_capacity(80);
+    loop {
+        line.clear();
+        let count = reader.read_line(&mut line)?;
+        if count == 0 {
+            return Err(crate::GetTimezoneError::FailedParsingString);
+        } else if line.starts_with("TZ=") {
+            line.truncate(line.trim_end().len());
+            line.replace_range(..3, "");
+            return Ok(line);
+        }
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -52,6 +52,7 @@ mod ffi_utils;
     any(target_os = "illumos", target_os = "solaris"),
     path = "tz_illumos.rs"
 )]
+#[cfg_attr(target_os = "aix", path = "tz_aix.rs")]
 #[cfg_attr(target_os = "android", path = "tz_android.rs")]
 #[cfg_attr(target_os = "haiku", path = "tz_haiku.rs")]
 mod platform;
--- a/src/platform.rs
+++ b/src/platform.rs
@@ -5,5 +5,5 @@ pub fn get_timezone_inner() -> std::result::Result<String, crate::GetTimezoneErr
 #[cfg(not(feature = "fallback"))]
 compile_error!(
     "iana-time-zone is currently implemented for Linux, Window, MacOS, FreeBSD, NetBSD, \
-    OpenBSD, Dragonfly, WebAssembly (browser), iOS, Illumos, Android, Solaris and Haiku.",
+    OpenBSD, Dragonfly, WebAssembly (browser), iOS, Illumos, Android, AIX, Solaris and Haiku.",
 );


Kijewski:
According to <https://www.ibm.com/docs/en/aix/7.1?topic=files-environment-file> and <https://unixhealthcheck.com/blog?id=291> AIX has its own time zone naming scheme.

The other implementations don't evaluate the environment variables by design. I would rather scan `/etc/environment` for `TZ=…` like in the [Solaris/Illumos implementation](https://github.com/strawlab/iana-time-zone/blob/41c9d380075420085951c1f63b428e6123f1e957/src/tz_illumos.rs).

ecnelises:
Hi @Kijewski , do you mean that the `(timezone, offset, daylight timezone, offset)` form should be transformed into `Region/City` form?

If so, mapping from `(timezone, offset)` to city is a one-to-many relation. For example, `CST-8` means either `Asia/Chongqing` `Asia/Shanghai` or `Asia/Harbin`. So arbitrary will be chosen?

ecnelises:
Gentle ping.. any further comments?

Kijewski:
I would scan `/etc/environment` for `TZ=…`, same as in illumos. This way the logic is the same as in our other platform implementations.

On AIX you can use Olson and POSIX time zones (<https://developer.ibm.com/articles/au-aix-posix/>). For no obvious reason the American POSIX time zones like [PST8PDT](https://en.wikipedia.org/wiki/PST8PDT) are also valid IANA time zone identifiers, but e.g. "CET-1CEST" is not.

I guess we can only hope that the system time zone is an IANA identifier. I would not like to add a white list. We could filter out results like `CET-1CEST-2,M3.5.0/02:00:00,M10.5.0/03:00:00` (anything that contains a comma), strip off everything after the comma, or let the caller decide how to use the information. 

@astraw, what do you think?

astraw:
Why is this still a draft?

I'm far from an AIX expert nor IANA database expert, but this all seems OK to me.

Although new to me, as far as I understand https://www.ietf.org/timezones/tzdb-2018g/theory.html, POSIX time zones are valid Olson `tz` names, so I don't think we need to do any filtering or parsing.

astraw (approved):

astraw:
@Kijewski do think the logic in my comment above is acceptable and we can merge this?

Kijewski:
As far as I understand it, every name in the Olson database is a valid value for a POSIX tz string, but the reverse is not true.

I am only mildly hesitant if we should allow POSIX tz strings, because a downstream consumer might assume that we are only returning valid names, and try to put it in a path name. E.g. in windows a colon can be used to for [file stream selection](https://stackoverflow.com/a/77240895/416224), but with AIX being a Unix implementation, I guess that a colon does not pose a security risk. In Unix colons are sometimes used as to separate paths, e.g. in the $PATH variable, but I don't think anyone would/should paste the their time zone in such a variable anyway.

And we aren't checking if the result is sane for any other OS, so I guess this PR is at least good enough. If the implementation can be made better in the future, then we can still update it.

astraw:
These are valid concerns and perhaps we should have a sanity check on the output shared across all implementations. I don't think that's a specific issue with this AIX implementation. Therefore I'll go ahead and merge this now and we can consider sanity checks in a platform independent way.

ecnelises:
Thanks!

src/tz_aix.rs
```diff
@@ -0,0 +1,26 @@
...
+        let count = reader.read_line(&mut line)?;
+        if count == 0 {
+            return Err(crate::GetTimezoneError::FailedParsingString);
+        } else if line.starts_with("TZ=") {
+            line.truncate(line.trim_end().len());
+            line.replace_range(..3, "");
+            return Ok(line);
+        }
+    }
+}
```
lopopolo, line 26: this function is unused.

src/tz_aix.rs
```diff
@@ -0,0 +1,26 @@
+use std::fs::OpenOptions;
+use std::io::{BufRead, BufReader};
+use std::env;
+
+pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
+    env::var("TZ").map_err(|_| crate::GetTimezoneError::OsError)
```
lopopolo, line 6: this will panic if `TZ` is not valid UTF-8.

this instead should do something like:

```rust
let timezone = env::var_os("TZ").ok_or(crate::GetTimezoneError::OsError)?;
timezone.into_string().map_err(|_| crate::GetTimezoneError::OsError)
```

lopopolo (commented):
hey @Kijewski @astraw I stumbled across this code and I think it could use a revisit

src/tz_aix.rs
astraw: According to the docs, [`std::env::var`](https://doc.rust-lang.org/stable/std/env/fn.var.html) does not panic on a Unicode error. (And it looks like this has been the case since rust 1.0.0.)

lopopolo: oh wow, I've had this wrong in my head for quite some time then. Sorry!

PR #58, opened 2022-09-12
Downgrade requirements for WASM dependencies

djc:
WebAssembly support was added in #38 targeting the current versions versions of wasm-bindgen and js-sys at the time, which is reasonable. Unfortunately the whoami crate, in trying to support older Rust versions, has reduced the upper version of wasm-bindgen that it supports, see https://github.com/ardaku/whoami/issues/43.

This PR thus reduces the minimum version required for wasm-bindgen. I used `cargo +nightly update -Z minimal-versions` with `cargo build --target wasm32-unknown-unknown` to find the earliest version that compiles. We could theoretically set up checks in CI to verify this on an ongoing basis, but this might be overkill for now.

cc @Kijewski

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,8 +25,8 @@ core-foundation-sys = "0.8.3"
 winapi = { version = "0.3.9", features = ["activation", "combaseapi", "objbase", "roapi", "winerror", "winstring"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-js-sys = "0.3.58"
-wasm-bindgen = "0.2.81"
+js-sys = "0.3.46"
+wasm-bindgen = "0.2.70"
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
-wasm-bindgen-test = "0.3.31"
+wasm-bindgen-test = "0.3"


astraw:
Rolling back the versions in Cargo.toml seems reasonable to me if you tested things work with those versions. I'm +1. @Kijewski what do you think?

(Adding CI tests to prove that the minimum package versions specified in Cargo.toml do function is a good idea - for all crates - but IMO not necessary at this stage. I haven't seen an implementation of such, nor have I looked.)

djc:
> Rolling back the versions in Cargo.toml seems reasonable to me if you tested things work with those versions.

Note that I only compile-tested one target (as mentioned in the description), I'm not normally a WASM user and running the tests against a WASM target didn't seem to work.

> (Adding CI tests to prove that the minimum package versions specified in Cargo.toml do function is a good idea - for all crates - but IMO not necessary at this stage. I haven't seen an implementation of such, nor have I looked.)

I think a number of crates ([like Tokio](https://github.com/tokio-rs/tokio/blob/master/.github/workflows/ci.yml#L306)) have started to exercise `cargo update -Z minimal-versions` in CI, but there are still a number of pain points to be ironed out (which is why Cargo hasn't stabilized `minimal-versions` yet).

Kijewski:
Thank you! The PR looks good to me, and I'll try it this evening.

It is possible to use `-Zminimal-versions` in `stable`, and also quite old versions: [ci.yml]. The documentation for [`RUSTC_BOOTSTRAP`] is a bit hidden, and you are not supposed to use it, but it works perfectly fine for use cases like this.

[ci.yml]: https://github.com/Kijewski/utcnow/blob/8f57fd17bf3e18ccaae79fdeef2de027ed0da6b3/.github/workflows/ci.yml#L69-L106
[`RUSTC_BOOTSTRAP`]: https://rustc-dev-guide.rust-lang.org/building/bootstrapping.html

Kijewski (approved):

astraw:
This is published in release 0.1.48.

PR #62, opened 2022-09-23
Follow up to #61

Kijewski:
* [`CStr::from_bytes_with_nul_unchecked()`] is only `const` since rustc 1.59.0. Just don't use it `const`.
* Moving the `static mut` variable into a getter function makes it more obvious that it cannot be changed after the initial assignment.

[`CStr::from_bytes_with_nul_unchecked()`]: https://doc.rust-lang.org/1.64.0/std/ffi/struct.CStr.html#method.from_bytes_with_nul_unchecked

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.50] - 2022-09-23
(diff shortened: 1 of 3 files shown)


astraw:
Looks good to me, so I merged this. Do you think it is worth a new release?

Kijewski:
Thank you!

I'd say yes, because I might have broken the workflow of some Android developer who compiles the code on Debian old-stable for whatever reason. But I think we can wait a few hours if there are more comments in here or #61. I'll wait until midnight CEST, and do the release then, so the date in the changelog is still correct. :)

Kijewski:
Published as 0.1.50.

Issue #64, opened 2022-09-25
Unsafe blocks review


pinkforest:
Hi considering this crate is getting popular :partying_face: congrats great work ~80k a day from less than a month :)

Just an idea .. 

or FWIW considering I am deeply interested about `unsafe` across the whole Rust :crab: ecosystem I wanted to chime in :)

Have you considered / or had anyone to vet these unsafe blocks in this crate properly through externally like some neutral 3rd party not deep in the project / associated crates ? I didn't see anything in [crev](https://web.crev.dev/rust-reviews/crates/) or [safety-dance](https://github.com/rust-secure-code/safety-dance).

You could also just hold back merging anything that touches any `unsafe {}` and put a label "Unsafe-Proposed" and let it sit for a week unless it is critical change for the community to chime in - that's what I do in advisory-db to radiate my intent - https://medium.com/@ElizAyer/dont-ask-forgiveness-radiate-intent-d36fd22393a3.

By radiating intent everyone feels having a chance to say something that leads less to haphazard changes.

(I happen to maintain [cargo-geiger](https://crates.io/crates/cargo-geiger) btw)

astraw:
Thanks for the thoughts and pointers.

I would be happy for someone to vet the unsafe blocks. How would you suggest we facilitate that?

The `Unsafe-Proposed` label also seems like a good idea.

We have become so "low level" that in some cases we cannot rely on higher level libraries, for example #50 and #61.

pinkforest:
We can organise this via safety-dance - I've added issue at https://github.com/rust-secure-code/safety-dance/issues/79

lopopolo:
One bit that I've been thinking about is that the windows and macOS internals are implemented by big `unsafe fn`. I don't think that's quite right. The functions are not unsafe to call (there's no invariants callers must uphold).

Instead, the functions seem to be declared as `unsafe fn` to avoid having to use and document `unsafe` blocks in their internals.

lopopolo:
I would like to find find a way to fail CI when the following lints fire:

- `unsafe_op_in_unsafe_fn`
- `clippy::undocumented_unsafe_blocks`

pinkforest:
For clippy you would have to add it into github action - that undocumented lint was stabilised *I think*

unsafe_op I am not sure it was stabilised so might need +nightly for cargo clippy

lopopolo:
@pinkforest ahh I meant getting the code changed so the lints would pass.

[...]

PR #65, opened 2022-09-27
macOS unsafe refactor

astraw:
As discussed in #64, especially https://github.com/strawlab/iana-time-zone/issues/64#issuecomment-1258944132, this refactoring to isolate only the specific calls that need to be unsafe.

I also created a new label [Unsafe-Proposed](https://github.com/strawlab/iana-time-zone/labels/Unsafe-Proposed) and assigned this PR with it. (Also as discussed in #64).

### DIFF
--- a/src/tz_macos.rs
+++ b/src/tz_macos.rs
@@ -20,12 +20,22 @@ fn get_timezone() -> Option<String> {
         return None;
     }
 
-    // If the name is encoded in UTF-8, copy it directly.
-    let cstr = unsafe { CFStringGetCStringPtr(name, kCFStringEncodingUTF8) };
-    if !cstr.is_null() {
-        let cstr = unsafe { std::ffi::CStr::from_ptr(cstr) };
-        if let Ok(name) = cstr.to_str() {
-            return Some(name.to_owned());
+    {
+        // New block prevents any variables escaping this scope.
+
+        // If the name is encoded in UTF-8, copy it directly.
+        let cstr_ptr = unsafe { CFStringGetCStringPtr(name, kCFStringEncodingUTF8) };
+        if !cstr_ptr.is_null() {
+            // Safety: 1) `cstr_ptr` must contain valid nul terminator. 2)
+            // `cstr_ptr` must be valid for reads of bytes up to and including
+            // the null terminator. 3) `cstr_ptr` must remain valid for lifetime
+            // of `cstr`. We assume #1 and #2 are true based on the result from
+            // CFStringGetCStringPtr. We ensure #3 by cloning the underlying &str
+            // to a String.
+            let cstr = unsafe { std::ffi::CStr::from_ptr(cstr_ptr) };
+            if let Ok(name) = cstr.to_str() {
+                return Some(name.to_owned());
+            }
         }
     }
 
--- a/src/tz_macos.rs
+++ b/src/tz_macos.rs
@@ -14,6 +14,8 @@ fn get_timezone() -> Option<String> {
     const MAX_LEN: usize = 32;
 
     // Get system time zone, and borrow its name.
+
+    // SAFETY: `Dropping::new()` would have returned `None` if the call to `CFTimeZoneCopySystem()` had failed.
     let tz = Dropping::new(unsafe { CFTimeZoneCopySystem() })?;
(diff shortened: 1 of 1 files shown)


src/tz_macos.rs
```diff
@@ -70,7 +82,7 @@ impl<T> Drop for Dropping<T> {
 
 impl<T> Dropping<T> {
     #[inline]
-    unsafe fn new(v: *const T) -> Option<Self> {
+    fn new(v: *const T) -> Option<Self> {
```
Kijewski, line 85: This method must remain `unsafe`. You can use only safe functions to construct an invalid pointer.

src/tz_macos.rs
```diff
@@ -34,18 +44,20 @@ unsafe fn get_timezone() -> Option<String> {
     let mut buf_bytes = 0;
     let range = CFRange {
         location: 0,
-        length: CFStringGetLength(name),
+        length: unsafe { CFStringGetLength(name) },
```
Kijewski, line 47:
```diff
+// SAFETY: we tested that `name` is not null, so we know that the result of `CFTimeZoneGetName()` is valid
```

src/tz_macos.rs
```diff
@@ -5,27 +5,37 @@ use core_foundation_sys::string::{
...
-    let cstr = CFStringGetCStringPtr(name, kCFStringEncodingUTF8);
-    if !cstr.is_null() {
-        let cstr = std::ffi::CStr::from_ptr(cstr);
-        if let Ok(name) = cstr.to_str() {
-            return Some(name.to_owned());
+    {
+        // New block prevents any variables escaping this scope.
+
+        // If the name is encoded in UTF-8, copy it directly.
+        let cstr_ptr = unsafe { CFStringGetCStringPtr(name, kCFStringEncodingUTF8) };
```
Kijewski, line 27:
```diff
+// SAFETY: we tested that `name` is not null, so we know that the result of `CFTimeZoneGetName()` is valid
```

src/tz_macos.rs
```diff
@@ -5,27 +5,37 @@ use core_foundation_sys::string::{
...
-unsafe fn get_timezone() -> Option<String> {
+fn get_timezone() -> Option<String> {
     // The longest name in the IANA time zone database is 25 ASCII characters long.
     const MAX_LEN: usize = 32;
 
     // Get system time zone, and borrow its name.
-    let tz = Dropping::new(CFTimeZoneCopySystem())?;
-    let name = CFTimeZoneGetName(tz.0);
+    let tz = Dropping::new(unsafe { CFTimeZoneCopySystem() })?;
+    let name = unsafe { CFTimeZoneGetName(tz.0) };
```
Kijewski, line 18:
```diff
+// SAFETY: `Dropping::new()` would have returned `None` if the call to `CFTimeZoneCopySystem()` had failed.
```

Kijewski (changes requested):

src/tz_macos.rs
lopopolo: Although I can see your argument that `Dropping::new` could be needed to be marked unsafe because the type will deterrence the pointer on drop.

src/tz_macos.rs
lopopolo: 👍 to @Kijewski. Let's add safety comments to these unsafe blocks. A safety comment should document how the code is upholding the invariants of the unsafe functions being called.

Kijewski:
Maybe a lot of things could be made more easy to read if the call of `CFTimeZoneCopySystem()` happened inside of `Dropping::new()`. (Dropping should get a new name then, though.)

lopopolo:
> I refactored the file a lot, and moved all the unsafe function calls into safe methods: https://github.com/Kijewski/iana-time-zone/blob/pr-macos-safe2/src/tz_macos.rs (totally untested).

Ooo that's nice! I like what you've done there @Kijewski. Please put up a PR. It'd make it easier to compare both approaches, give them reviews, and see what we like better.

lopopolo:
I also want to note that if we're interested, we could ask the Servo folks again to merge this change which would allow us to use the higher level bindings again:

- https://github.com/servo/core-foundation-rs/pull/521

astraw:
I like #67 more than this one. I am closing this and let's move the discussion there.

PR #66, opened 2022-09-28
Implement for Haiku 

Kijewski:
Tested on Haiku R1/beta3.

The stress-test takes 1m43s. That's quite terrible.


lopopolo:
Is binding to the haiku C++ API with a small native extension compiled by `cc` in `build.rs` an option here? This looks like it would be a lot simpler and less error prone since it is one API call.

https://www.haiku-os.org/docs/api/classBTimeZone.html

Kijewski:
Sure! Do you have experience how to work with `cxx`?

[`BLocaleRoster::GetDefaultTimeZone()`](https://www.haiku-os.org/docs/api/classBLocaleRoster.html#a7cb5f911107993a9c4816b52e9221572) + [`BTimeZone::ID()`](https://www.haiku-os.org/docs/api/classBTimeZone.html#afa38541edc05b13c25618f1bca75c6b0) + [`BString::CopyInto(&buf, 0, i32::MAX)`](https://github.com/haiku/haiku/blob/530f89aa6d0a551aa8ab4a3af16d88186f04878c/src/kits/support/String.cpp#L471) should do exactly what we need.

lopopolo:
I don't have experience with ccx to call this directly in Rust, but a tiny `.cpp` file with a C ABI export and one header can be made pretty usable with the `cc` crate and bindgen in a `build.rs` script

Kijewski:
It works. The stress-test now executes in 19 seconds.

Well, at least on an x86_64 system it works. But even if it only works for 64bit systems, I guess that's good enough.

lopopolo:
This is looking great!

If you'd like to add clang-format to CI, we can use: https://github.com/artichoke/clang-format. I use it in Artichoke for C code. I recently added C++ support to the file scanner.

I tried it out with a simple `.clang-format` config and came up with this:

```diff
commit 20fa465a054534b98955a721ad5ba7dad03e851a (HEAD -> pr-haiku)
Author: Ryan Lopopolo <rjl@hyperbo.la>
Date:   Wed Sep 28 19:41:25 2022 -0700

    Add clang-format config

diff --git a/.clang-format b/.clang-format
new file mode 100644
index 0000000..444c17f
--- /dev/null
+++ b/.clang-format
@@ -0,0 +1,9 @@
+---
+BasedOnStyle: Google
+IndentWidth: 4
+ColumnLimit: 100
+---
+Language: Cpp
+SortIncludes: false
+TabWidth: 4
+UseTab: Never
diff --git a/include/impl_haiku.h b/include/impl_haiku.h
index cd50fe8..2d9c362 100644
--- a/include/impl_haiku.h
+++ b/include/impl_haiku.h
@@ -5,5 +5,5 @@
 #include <cstddef>

 namespace tz_haiku {
-    size_t get_tz(uint8_t *buf, size_t buf_len);
+size_t get_tz(uint8_t *buf, size_t buf_len);
 }
```

A GitHub Actions workflow to check this in CI looks like:

```yaml
  c:
    name: Lint and format C
    runs-on: ubuntu-latest
    steps:
      - name: Checkout repository
        uses: actions/checkout@v3

      - name: Setup Node.js runtime
        uses: actions/setup-node@v3
        with:
          node-version: 16

      - name: Install npm
        run: npm i -f -g npm@8.16.0

      - name: Lint and check formatting with clang-format
        run: npx github:artichoke/clang-format --check
```

src/tz_haiku.rs
```diff
@@ -0,0 +1,23 @@
...
+    let mut buf = [0u8; 32];
+    // SAFETY: the buffer is valid and called function ensures that sizeof(u8) == sizeof(c_char)
+    let len = unsafe { ffi::get_tz(buf.as_mut_ptr(), buf.len()) };
+    // The name should not be empty, or excessively long.
+    if len > 0 && len < buf.len() {
+        let s = std::str::from_utf8(&buf[..len]).map_err(|_| crate::GetTimezoneError::OsError)?;
+        Ok(s.to_owned())
+    } else {
+        Err(crate::GetTimezoneError::OsError)
+    }
```
lopopolo, line 22: I think this reads a bit clearer and avoids doing the length check _and_ a panicking index operation.

```suggestion
    match buf.get(..len) {
        Some(s) if s.is_empty() => Err(crate::GetTimezoneError::OsError),
        Some(s) => {
            let s = std::str::from_utf8(s).map_err(|_| crate::GetTimezoneError::OsError)?;
            Ok(s.to_owned())
        }
        None => Err(crate::GetTimezoneError::OsError),
    }
```

lopopolo (changes requested):
I left a suggested change to make the Rust side of things a bit easier to read and I was hoping we could add some comments in a couple of places on the C++ side of things.

Thanks for digging into cxx and the Haiku C++ docs. I like this a lot better than the initial pass at this PR.

src/tz_haiku.rs
```diff
@@ -0,0 +1,23 @@
+#[cxx::bridge(namespace = "tz_haiku")]
+mod ffi {
+    unsafe extern "C++" {
+        include!("iana-time-zone/include/impl_haiku.h");
+
+        /// SAFETY: `buf` must be valid and be at least `buf_len` bytes long
+        unsafe fn get_tz(buf: *mut u8, buf_len: usize) -> usize;
```
Kijewski, line 7: This function can be implemented much simpler without using `unsafe` or casting slices to pointers manually: https://cxx.rs/binding/slice.html. I did not notice this feature yesterday. Will change it this evening or tomorrow.

src/impl_haiku.cc
```diff
@@ -0,0 +1,40 @@
...
+        return 0;
+    }
+
+    BString bname(tz.ID());
+    int32_t length(bname.Length());
+    if (length <= 0 || size_t(length) > buf_len) {
+        return 0;
+    }
+
+    const char *sname(bname.String());
```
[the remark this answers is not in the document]
Kijewski: Yes, the string is borrowed. I'll add a comment.

> You can't modify or free the pointer. Once the [BString](https://www.haiku-os.org/docs/api/classBString.html) object is deleted, the pointer becomes invalid.

https://www.haiku-os.org/docs/api/classBString.html#ae4fe78b06c8e3310093b80305e14ba87  
https://github.com/haiku/haiku/blob/530f89aa6d0a551aa8ab4a3af16d88186f04878c/headers/os/support/String.h#L464-L470

src/impl_haiku.cc
```diff
@@ -0,0 +1,40 @@
+#include "iana-time-zone/include/impl_haiku.h"
+#include "iana-time-zone/src/tz_haiku.rs.h"
+
+#include <cstring>
+#include <Errors.h>
+#include <LocaleRoster.h>
+#include <String.h>
+#include <TimeZone.h>
+
+size_t ::tz_haiku::get_tz(uint8_t *buf, size_t buf_len) {
```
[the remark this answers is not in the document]
Kijewski: Haiku only declares their `malloc` / `realloc` / `free` / `operator new` functions as `noexcept`. On amd64's ABI a `try` block should have almost no overhead, if there was no exception thrown. Putting everything into a `try` block seems like a good idea.

Kijewski:
I don't understand how this change could break the wasm implementation?? https://github.com/strawlab/iana-time-zone/actions/runs/3148078221/jobs/5118236934

src/impl_haiku.cc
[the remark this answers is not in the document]
lopopolo: thanks for adding this.

include/impl_haiku.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include "rust/cxx.h"
+
+#include <cstddef>
+
+namespace tz_haiku {
+    size_t get_tz(uint8_t *buf, size_t buf_len);
```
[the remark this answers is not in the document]
lopopolo: this was addressed by using `rust::Slice` type from `cxx`.

src/tz_haiku.rs
lopopolo: done in https://github.com/strawlab/iana-time-zone/pull/66/commits/742f223624a46326a7e4d304d4894c6c38950742

lopopolo:
I'm not sure either, but I bumped the minimum version of `js-sys` which seems to have fixed it for me locally

Cargo.toml
```diff
@@ -24,8 +24,14 @@ core-foundation-sys = "0.8.3"
 winapi = { version = "0.3.9", features = ["activation", "combaseapi", "objbase", "roapi", "winerror", "winstring"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-js-sys = "0.3.46"
+js-sys = "0.3.50"
```
lopopolo, line 27: if I had to take a guess at why this is required, `cxx` probably has a different minimum version of `syn` or some other proc macro crate which doesn't play nice with earlier versions of `js-sys`.

src/impl_haiku.cc
```diff
@@ -0,0 +1,46 @@
...
+
+        // BString::String() returns a borrowed string.
+        // https://www.haiku-os.org/docs/api/classBString.html#ae4fe78b06c8e3310093b80305e14ba87
+        const char *sname(bname.String());
+        if (!sname) {
+            return 0;
+        }
+
+        std::memcpy(buf.data(), sname, size_t(length));
+        return length;
```
lopopolo, line 42: we're relying on integer promotion rules here and also cast `length` to `size_t` in several places.

src/impl_haiku.cc
```diff
@@ -0,0 +1,46 @@
...
+            return 0;
+        }
+
+        BTimeZone tz(NULL, NULL);
+        if (locale_roster->GetDefaultTimeZone(&tz) != B_OK) {
+            return 0;
+        }
+
+        BString bname(tz.ID());
+        int32_t length(bname.Length());
```
lopopolo, line 29: I don't know if this actually compiles, but can we do something like this:

```suggestion
        size_t length(size_t(bname.Length()));
```

lopopolo (changes requested):

lopopolo:
This is looking good! One more question on the C++ side.

src/impl_haiku.cc
lopopolo: ahh maybe it's not that simple since we check if `length < 0` just beyond this. This is a spot where I prefer Rust shadowing ability.

Cargo.toml
Kijewski: Thanks. Yes, something like that seems likely.

src/impl_haiku.cc
Kijewski: I introduced a new variable after checking that it's not negative. Yes, rust's variable shadowing is a very nice feature that I guess would work well with C++'s scoping rules, too. But I don't think that C++ will ever adopt this feature. Otherwise code like this wouldn't compile anymore: https://godbolt.org/z/8s8zM9K6h

```cpp
int main(int, char**) {
    int result(result);
    return result;
}
```

lopopolo (approved):

lopopolo:
@astraw this looks good to me. Do you want to take a quick look before we start the 7 day clock on this unsafe-proposed PR?

astraw:
Thanks.

However, I think a `build.rs` on all platforms is a heavy price to pay to support a single platform, namely haiku. Can we move the haiku specific build.rs and cxx stuff to a separate crate (in the same git repo) so that `build.rs` is isolated to that platform?

Kijewski:
build.rs's main function is empty for other OSes. `cxx` is only included for that one OS, because everything is wrapped in

```rust
    #[cfg(target_os = "haiku")]
    let _: () = {
        ....
    }
```


I don't think that having an empty build script would make a noticeably difference.

astraw:
The concern is that the build script must be executed (not what it executes per se). I'm not so much worried about execution time etc but rather about cross compilation or using other toolchains etc.

Kijewski:
Cross compilation is not affected, because the build script is always only executed on the host. At least I don't see how it could be affected right now.

E.g. if your build.rs to read `/etc/hostname` that will work on a linux host even when cross compiling to windows. But it will fail on a windows host, even when cross compiling to linux.

astraw:
Yes, a `build.rs` file triggers building and execution of code on the host during cross compilation. My point is that this brings  complexity for no gain on all platforms except Haiku. To me, the alternative is relatively painless and only paid by people building for Haiku.

Kijewski:
Yeah, also this way users of other OS don't need to wonder why iana-time-zone executes a build script all of a sudden. I split off the code into its own crate.

haiku/src/implementation.cc
```diff
@@ -0,0 +1,60 @@
...
+
+size_t ::iana_time_zone_haiku::get_tz(rust::Slice<uint8_t> buf) {
+    try {
+        static_assert(sizeof(char) == sizeof(uint8_t), "Illegal char size");
+
+        if (buf.empty()) {
+            return 0;
+        }
+
+        BLocaleRoster *locale_roster(BLocaleRoster::Default());
```
astraw, line 21: Can you add a comment here saying that `BLocaleRoster::Default()` returns a static pointer and thus no free is necessary?

astraw (changes requested):
Apart from the one trivial change, this looks good to me.

haiku/src/implementation.cc
Kijewski: The [documentation](https://www.haiku-os.org/docs/api/classBLocaleRoster.html#a7cb5f911107993a9c4816b52e9221572) is not terribly explicit about what to do with the returned pointer, but at the place of the implementation [a comment](https://github.com/haiku/haiku/blob/8f16317a5b6db5c672f331814273e5857555020f/src/kits/locale/LocaleRoster.cpp#L143-L147) indicates that the pointer is static. → [`MutableLocaleRoster::Default()`](https://github.com/haiku/haiku/blob/8f16317a5b6db5c672f331814273e5857555020f/src/kits/locale/MutableLocaleRoster.cpp#L59-L66)

astraw (approved):

astraw:
We have 2 approvals now, so according to our `Unsafe-Proposed` policy (#64), let's let this sit until 8 October and then merge.

astraw:
@Kijewski thanks for the work on this, including shifting most of it to its own crate.

PR #67, opened 2022-09-29
Refactor MacOS implementation a lot

Kijewski:
Cf. #64, #65.

### DIFF
--- a/src/tz_macos.rs
+++ b/src/tz_macos.rs
@@ -122,11 +122,13 @@ mod string_ref {
                     &mut buf_bytes,
                 )
             };
-            if converted_bytes != length || buf_bytes < 0 || buf_bytes as usize > buf.len() {
-                None
-            } else {
-                std::str::from_utf8(&buf[..buf_bytes as usize]).ok()
+            if converted_bytes != length {
+                return None;
             }
+            use std::convert::TryFrom;
+            let len = usize::try_from(buf_bytes).ok()?;
+            let s = buf.get(..len)?;
+            std::str::from_utf8(s).ok()
         }
     }
 }
--- a/src/tz_macos.rs
+++ b/src/tz_macos.rs
@@ -60,7 +60,8 @@ mod system_time_zone {
             if string.is_null() {
                 None
             } else {
-                Some(super::string_ref::StringRef::new(string, self))
+                // SAFETY: here we ensure that `string` is a valid pointer.
+                Some(unsafe { super::string_ref::StringRef::new(string, self) })
             }
         }
     }
@@ -83,7 +84,8 @@ mod string_ref {
     }
 
     impl<'a, T> StringRef<'a, T> {
-        pub(crate) fn new(string: CFStringRef, _parent: &'a T) -> Self {
+        // SAFETY: `StringRef` must be valid pointer
+        pub(crate) unsafe fn new(string: CFStringRef, _parent: &'a T) -> Self {
             Self { string, _parent }
         }
 
--- a/src/tz_macos.rs
+++ b/src/tz_macos.rs
@@ -30,6 +30,7 @@ fn get_timezone() -> Option<String> {
 
 mod system_time_zone {
     //! create a safe wrapper around `CFTimeZoneRef`
+
     use core_foundation_sys::base::{CFRelease, CFTypeRef};
     use core_foundation_sys::timezone::{CFTimeZoneCopySystem, CFTimeZoneGetName, CFTimeZoneRef};
 
@@ -68,12 +69,12 @@ mod system_time_zone {
 mod string_ref {
     //! create safe wrapper around `CFStringRef`
 
-    use core_foundation_sys::{
-        base::{Boolean, CFRange},
-        string::{
-            kCFStringEncodingUTF8, CFStringGetBytes, CFStringGetCStringPtr, CFStringGetLength,
-            CFStringRef,
-        },
+    use std::convert::TryInto;
+
+    use core_foundation_sys::base::{Boolean, CFRange};
+    use core_foundation_sys::string::{
+        kCFStringEncodingUTF8, CFStringGetBytes, CFStringGetCStringPtr, CFStringGetLength,
+        CFStringRef,
     };
 
     pub(crate) struct StringRef<'a, T> {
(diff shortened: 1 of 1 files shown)


astraw:
@Kijewski I thought it was easier to read and safer if I moved each newtype to its own module. Along the way I did a small bit of refactoring. I have some concern this will not meet our MSRV requirements so may need to be updated once the CI tests are done. But it does actually run correctly on my Mac :)

src/tz_macos.rs
```diff
@@ -1,80 +1,132 @@
...
+                    buf.as_mut_ptr(),
+                    buf.len() as isize,
+                    &mut buf_bytes,
+                )
+            };
+            if converted_bytes != length || buf_bytes < 0 || buf_bytes as usize > buf.len() {
+                None
+            } else {
+                std::str::from_utf8(&buf[..buf_bytes as usize]).ok()
+            }
```
lopopolo, line 129: This breaks up the conditional a bit and removes the duplicate len check and panicking index operation.

```suggestion
            if converted_bytes != length {
                return None;
            }
            let len = usize::try_from(buf_bytes).ok()?;
            let s = buf.get(..len)?;
            std::str::from_utf8(s).ok()
```

lopopolo (approved):
This LGTM. I left one proposed tweak to break apart a gnarly conditional.

src/tz_macos.rs
lopopolo: Probably need a `use std::convert::TryFrom;` to make this work.

astraw: thanks, yes, this is much nicer. Pushed in 55ddac33aca8aa4e286c05d16d9b3478995cb6c5. Thanks @lopopolo

astraw: Drat, `std::convert::TryFrom` was introduced in rust 1.34, whereas our MSRV is 1.31, so I had to revert this.

astraw:
This looks good to me. Thanks @Kijewski for the main work here. And the CI tests do all pass.

It seems we have a consensus amongst committers/reviewers that this is ready to merge. However, as it carries the `Unsafe-Proposed` label, let's let this sit for a week (so, until 6 Oct) to attract review as discussed in #64. Also, if someone is auditing this crate (e.g. https://github.com/rust-secure-code/safety-dance/issues/79 ), I ask them to please check this implementation as an alternative to the existing macOS implementation.

lopopolo:
Thanks all! This looks like a big improvement.

src/tz_macos.rs
lopopolo: Ahh that's unfortunate.

`chrono` 0.4.x's MSRV is currently 1.38. Do you think this would be worth considering an MSRV bump to 1.34?

Kijewski:
> I thought it was easier to read and safer if I moved each newtype to its own module.

I like that very much, @astraw. Now the type are truly self contained.

src/tz_macos.rs
lopopolo: we bumped MSRV in #70 which allowed us to make this change.

src/tz_macos.rs
```diff
@@ -1,80 +1,135 @@
...
+
+    use core_foundation_sys::base::{Boolean, CFRange};
+    use core_foundation_sys::string::{
+        kCFStringEncodingUTF8, CFStringGetBytes, CFStringGetCStringPtr, CFStringGetLength,
+        CFStringRef,
+    };
+
+    pub(crate) struct StringRef<'a, T> {
+        string: CFStringRef,
+        _parent: &'a T,
```
bjorn3, line 82: This can be `PhantomData<&'a ()>`, right?

src/tz_macos.rs
```diff
@@ -1,80 +1,135 @@
...
+    }
+
+    impl<'a, T> StringRef<'a, T> {
+        pub(crate) fn new(string: CFStringRef, _parent: &'a T) -> Self {
```
bjorn3, line 86: This function is unsafe as it requires the passed in pointer to be valid.

src/tz_macos.rs
```diff
@@ -1,80 +1,135 @@
...
+            // SAFETY: No invariants to uphold. We'll release the pointer when we don't need it anymore.
+            let v: CFTimeZoneRef = unsafe { CFTimeZoneCopySystem() };
+            if v.is_null() {
+                None
+            } else {
+                Some(SystemTimeZone(v))
+            }
+        }
+
+        pub(crate) fn name(&self) -> Option<super::string_ref::StringRef<Self>> {
```
bjorn3, line 57: Could you explicitly specify lifetime parameters here to clarify that the `StringRef` is bound to `&self`?

bjorn3 (approved):
I think this code doesn't have UB, but I do have a couple of suggestions to the safe code outside of `system_time_zone` and `string_ref` can't cause UB even when changing it.

src/tz_macos.rs
bjorn3: What if you pass `1 as CFStringRef`? Or `let foo = 0u8; &FOO as *const u8 as CFStringRef`? Both are not valid `CFStringRef`, yet can't be catched with a null pointer check or any check that doesn't involve something like valgrind for that matter.

src/tz_macos.rs
astraw: I'm not sure if that excludes mutable access to the parent and, perhaps more fundamentally, if that is an invariant we need to uphold?

For those reasons, the approach here (taking an actual reference to the original) seems more less unsafe but I am not certain of this.

src/tz_macos.rs
astraw: I marked the function as unsafe in 9fbf7a6d7aadc4840181b72b3c4997099be48008

src/tz_macos.rs
bjorn3: You're right. I now get how you exactly meant to enforce this. Feel free to ignore https://github.com/strawlab/iana-time-zone/pull/67/files#r988930534.

astraw: Thanks. I made a comment in the code indicating this now in 770edec2125540da62d1eabe351408b4014e27c3

src/tz_macos.rs
astraw: I added some documentation about this in 8b9c9f0af2b35e8a083a128f539997849241d294. Is that what you had in mind?

astraw:
The CI failures on the last commits here (`archive member 'rust.metadata.bin' with length 4847795 is not mach-o or llvm bitcode file`) seem unrelated to the changes made. They arise on macOS with rustc 1.38 but not current stable or nightly. For this reason I propose we ignore them, but I'm really not sure what is going on.

astraw:
The test failures are now documented in #72, which is unrelated.

Thanks everyone for the feedback and especially @Kijewski for the work on this. I hope your recovery is well underway.

PR #69, opened 2022-09-30
Refactor Android property key `CStr` construction to add tests

lopopolo:
Add a new FFI utils module that is always tested on all platforms.

Extract the C string bytes to a const and getter function, add more docs, add tests that the const can be safely turned into a `CStr`.

### DIFF
--- a/src/ffi_utils.rs
+++ b/src/ffi_utils.rs
@@ -1,0 +1,49 @@
+//! Cross platform FFI helpers.
+
+use std::ffi::CStr;
+
+// The system property named 'persist.sys.timezone' contains the name of the
+// current timezone.
+//
+// From https://android.googlesource.com/platform/bionic/+/gingerbread-release/libc/docs/OVERVIEW.TXT#79:
+//
+// > The name of the current timezone is taken from the TZ environment variable,
+// > if defined. Otherwise, the system property named 'persist.sys.timezone' is
+// > checked instead.
+const ANDROID_TIMEZONE_PROPERTY_NAME: &[u8] = b"persist.sys.timezone\0";
+
+/// Return a [`CStr`] to access the timezone from an Android system properties
+/// environment.
+pub fn android_timezone_property_name() -> &'static CStr {
+    // In tests or debug mode, opt into extra runtime checks.
+    if cfg!(any(test, debug_assertions)) {
+        return CStr::from_bytes_with_nul(ANDROID_TIMEZONE_PROPERTY_NAME).unwrap();
+    }
+
+    // SAFETY: the key is NUL-terminated and there are no other NULs, this
+    // invariant is checked in tests.
+    unsafe { CStr::from_bytes_with_nul_unchecked(ANDROID_TIMEZONE_PROPERTY_NAME) }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::ffi::CStr;
+
+    use super::{android_timezone_property_name, ANDROID_TIMEZONE_PROPERTY_NAME};
+
+    #[test]
+    fn test_android_timezone_property_name_is_valid_cstr() {
+        CStr::from_bytes_with_nul(ANDROID_TIMEZONE_PROPERTY_NAME).unwrap();
+
+        let mut invalid_property_name = ANDROID_TIMEZONE_PROPERTY_NAME.to_owned();
+        invalid_property_name.push(b'\0');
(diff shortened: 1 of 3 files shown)


astraw (approved):
Looks good. Thanks.

Kijewski:
I updated the CI to make sure the test is executed, because I don't anyone of us will won't to try every commit an our actual phones. :)

src/lib.rs
```diff
@@ -86,12 +86,10 @@ pub fn get_timezone() -> std::result::Result<String, crate::GetTimezoneError> {
     platform::get_timezone_inner()
 }
 
-#[cfg(test)]
-mod tests {
```
lopopolo, line 90: why did we unwrap this outside of the cfg'd `mod tests`?

.github/workflows/rust.yml
```diff
@@ -40,6 +40,38 @@ jobs:
       - run: cargo test --all-targets
       - run: cargo run --example stress-test
 
+  test-only:
```
lopopolo, line 43: instead of blowing up CI with even more checks, would y'all be open to moving the const and the test to a `ffi_utils.rs` mod or something that we keep platform agonstic?

src/lib.rs
Kijewski: My thinking was, that someone might add tests to this `mod` that don't get executed because of the `not(android)` guard, even though the tests might be useful on android, too. I.e. that the `not(android)` part was too easy to overlook. I guess that's overthinking it.

Probably it would have been even smarter to add the guard to the actual test and let the `mod` stay. I'll fix that. :)

.github/workflows/rust.yml
Kijewski: Yes, that would make sense.

lopopolo: done

lopopolo:
@astraw @Kijewski I took another stab at this to reduce some of the complexity in CI and add more tests.

Kijewski (approved):

src/ffi_utils.rs
```diff
@@ -0,0 +1,49 @@
...
+// From https://android.googlesource.com/platform/bionic/+/gingerbread-release/libc/docs/OVERVIEW.TXT#79:
+//
+// > The name of the current timezone is taken from the TZ environment variable,
+// > if defined. Otherwise, the system property named 'persist.sys.timezone' is
+// > checked instead.
+const ANDROID_TIMEZONE_PROPERTY_NAME: &[u8] = b"persist.sys.timezone\0";
+
+/// Return a [`CStr`] to access the timezone from an Android system properties
+/// environment.
+pub fn android_timezone_property_name() -> &'static CStr {
```
astraw, line 17: Potential optimization: the function signature can be made `const` if the runtime check branch inside it is removed. Since the tests are below, this seems like an easy win. That said, the compiler probably figures this out anyway. Moving the unsafe to explicitly compile-time also seems safer and could then be mentioned in the `SAFETY` docstring.

I think the function should be made `pub(crate)` to signal unambiguously this is not public API.

lopopolo: The CStr function is only const on recent Rust (1.59 or something, I'm on mobile and not looking at the docs)

astraw: Ah, I had missed that. You are right, we can't switch to `const` until we can use rust 1.59.

I think `pub(crate)` should be good, though.

Kijewski:
I added `Tier-` to `Tier-3` labels to match https://doc.rust-lang.org/nightly/rustc/platform-support.html. E.g Android is a Tier 2 platform.

src/ffi_utils.rs
lopopolo: done

lopopolo:
Alright we have two approvals, I've addressed all of the comments, and the build is green. Since this has the `Unsafe-Proposed` label, I'll merge this on October 7.

lopopolo:
> I added `Tier-` to `Tier-3` labels to match https://doc.rust-lang.org/nightly/rustc/platform-support.html. E.g Android is a Tier 2 platform.

Thanks for this! I went through the full history of PRs and issues and added these tags the best I could.

astraw:
Great, thanks @lopopolo .

This is now ready to merge. Given its `Unsafe-Proposed` status, the 7 day timer starts now. So we should merge this on 8 October.

astraw:
(@lopopolo somehow I missed your earlier message about merging on 7 October. Sorry about that!)

PR #75, opened 2022-10-09
Get Haiku platform crate compiling on non-Haiku platforms

lopopolo:
Some changes to get `iana-time-zone-haiku`'s fallback code compiling and working correctly on non-Haiku hosts:

- Only link to `be` lib when the target is a Haiku target triple.
- Parse the target triple in `build.rs` from the `TARGET` env var which is set by Cargo.
- Fix cargo pragmas in `build.rs` to properly emit `rerun-if-changed` directives.
- Use C++11 namespace declaration syntax (the current `implementation.cc` file used C++17 syntax which I imagine was not caught because `-std=c++11` is only passed if supported by the host cxx compiler).
- Add CI to build and test `iana-time-zone-haiku` on macOS, Linux, and Windows.
- Add CI to run `cargo check` and `cargo clippy` on `iana-time-zone-haiku`.
- Add standard set of lints to `iana-time-zone-haiku`. Extracted from #76.
- Add `Cargo.toml` keywords and categories to `iana-time-zone-haiku`.
- Fix typo where crate level docs were incorrectly declared with `///` (changed to `//!`)
- Add docs to `iana_time_zone_haiku::get_timezone()`.
- Add example to `iana_time_zone_haiku::get_timezone()`.
- Add smoke test for `iana_time_zone_haiku::get_timezone()` to ensure the fallback routine returns `None` on all non-Haiku platforms.
- Add smoke test for `iana_time_zone_haiku::get_timezone()` on Haiku to ensure `Some` is returned and that the resulting string is non-empty.

### DIFF
--- a/haiku/build.rs
+++ b/haiku/build.rs
@@ -6,9 +6,9 @@ fn main() {
         .flag_if_supported("-std=c++11")
         .compile("tz_haiku");
 
-    println!("cargo:return-if-changed=src/lib.rs");
-    println!("cargo:return-if-changed=src/implementation.cc");
-    println!("cargo:return-if-changed=src/interface.h");
+    println!("cargo:rerun-if-changed=src/lib.rs");
+    println!("cargo:rerun-if-changed=src/implementation.cc");
+    println!("cargo:rerun-if-changed=src/interface.h");
 
     let target = env::var_os("TARGET").expect("cargo should set TARGET env var");
     let target = target
--- a/haiku/build.rs
+++ b/haiku/build.rs
@@ -1,11 +1,20 @@
+use std::env;
+
 fn main() {
     cxx_build::bridge("src/lib.rs")
         .file("src/implementation.cc")
         .flag_if_supported("-std=c++11")
         .compile("tz_haiku");
 
-    println!("cargo:return-if-changed=include/interface.h");
-    println!("cargo:return-if-changed=src/implementation.cc");
     println!("cargo:return-if-changed=src/lib.rs");
-    println!("cargo:rustc-link-lib=be");
+    println!("cargo:return-if-changed=src/implementation.cc");
+    println!("cargo:return-if-changed=src/interface.h");
+
+    let target = env::var_os("TARGET").expect("cargo should set TARGET env var");
+    let target = target
+        .to_str()
+        .expect("TARGET env var should be valid UTF-8");
+    if target.contains("haiku") {
+        println!("cargo:rustc-link-lib=be");
+    }
 }
--- a/haiku/src/implementation.cc
+++ b/haiku/src/implementation.cc
@@ -1,9 +1,9 @@
 #include "iana-time-zone-haiku/src/interface.h"
 #include "iana-time-zone-haiku/src/lib.rs.h"
 
-#ifdef __HAIKU__
-
 #include <cstring>
+
+#ifdef __HAIKU__
 
 #include <Errors.h>
 #include <LocaleRoster.h>
@@ -57,6 +57,8 @@ size_t ::iana_time_zone_haiku::get_tz(rust::Slice<uint8_t> buf) {
 
 #else
 
-size_t ::iana_time_zone_haiku::get_tz(rust::Slice<uint8_t>) { return 0; }
+namespace iana_time_zone_haiku {
+size_t get_tz(rust::Slice<uint8_t>) { return 0; }
+}  // namespace iana_time_zone_haiku
 
 #endif
--- a/haiku/src/implementation.cc
+++ b/haiku/src/implementation.cc
@@ -1,9 +1,9 @@
 #include "iana-time-zone-haiku/src/interface.h"
 #include "iana-time-zone-haiku/src/lib.rs.h"
 
-#include <cstring>
-
 #ifdef __HAIKU__
+
+#include <cstring>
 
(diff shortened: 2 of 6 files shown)


haiku/src/implementation.cc
```diff
@@ -57,6 +57,8 @@ size_t ::iana_time_zone_haiku::get_tz(rust::Slice<uint8_t> buf) {
 
 #else
 
-size_t ::iana_time_zone_haiku::get_tz(rust::Slice<uint8_t>) { return 0; }
```
lopopolo, line 60: If I'm reading this set of docs correctly, this declaration syntax is only available since C++17: https://en.cppreference.com/w/cpp/language/namespace.

haiku/build.rs
```diff
@@ -1,11 +1,20 @@
...
-    println!("cargo:rustc-link-lib=be");
+    println!("cargo:return-if-changed=src/implementation.cc");
+    println!("cargo:return-if-changed=src/interface.h");
+
+    let target = env::var_os("TARGET").expect("cargo should set TARGET env var");
+    let target = target
+        .to_str()
+        .expect("TARGET env var should be valid UTF-8");
+    if target.contains("haiku") {
+        println!("cargo:rustc-link-lib=be");
```
lopopolo, line 18: only link to `be` lib on haiku cross compile targets.

Because the build script runs on the host, we must parse the `TARGET` env variable that cargo sets.

haiku/src/implementation.cc
Kijewski: You're right! Could you do the same in line 13?

haiku/src/implementation.cc
```diff
@@ -1,10 +1,10 @@
 #include "iana-time-zone-haiku/src/interface.h"
 #include "iana-time-zone-haiku/src/lib.rs.h"
 
-#ifdef __HAIKU__
-
 #include <cstring>
 
+#ifdef __HAIKU__
```
Kijewski, line 6: Why did you move `#include <cstring>` outside of the guarded block?

haiku/src/implementation.cc
lopopolo: done

haiku/src/implementation.cc
lopopolo: because `size_t` is defined there which is needed for the `#else` block.

lopopolo: gotcha. I undid these changes.

haiku/build.rs
```diff
@@ -1,11 +1,20 @@
...
         .file("src/implementation.cc")
         .flag_if_supported("-std=c++11")
         .compile("tz_haiku");
 
-    println!("cargo:return-if-changed=include/interface.h");
-    println!("cargo:return-if-changed=src/implementation.cc");
     println!("cargo:return-if-changed=src/lib.rs");
-    println!("cargo:rustc-link-lib=be");
+    println!("cargo:return-if-changed=src/implementation.cc");
+    println!("cargo:return-if-changed=src/interface.h");
```
Kijewski, line 11: Oops, thank you for catching that!

lopopolo: thanks for commenting here. The build.rs directive is _rerun_ if changed, not return if changed. I pushed a new commit.

Kijewski (approved):

haiku/build.rs
Kijewski: Strange, I thought I copied the line instead of typing it out of memory. I have no idea where the typo came from.

astraw:
@lopopolo this looks good.

Would it be possible to add configuration such that github actions does indeed run and build the haiku module, e.g. on macOS? Also, how does macOS actually factor in here? Is this the OS you developed this on?

lopopolo:
@astraw I pushed some CI changes and pulled forward the warnings work from #76 specific to the haiku platform crate, fixed up docs, added another test.

lopopolo:
I removed the `-Zminimal-versions` build for the haiku crate because `cxx-build` does not build with minimal versions. See:

- https://github.com/dtolnay/cxx/pull/1111

haiku/src/lib.rs
```diff
@@ -27,3 +50,19 @@ pub fn get_timezone() -> Option<String> {
...
+    #[test]
+    #[cfg(not(target_os = "haiku"))]
+    fn test_fallback_on_non_haiku_platforms() {
+        assert!(super::get_timezone().is_none());
+    }
+
+    #[test]
+    #[cfg(target_os = "haiku")]
+    fn test_retrieve_time_zone_on_haiku_platforms() {
+        let timezone = super::get_timezone().unwrap();
```
lopopolo, line 65: I'm not sure if we should keep this test since the function is technically (and documented to be) fallible.

lopopolo:
all the nightly android cross builds are failing. not sure why.

astraw (approved):

lopopolo:
@astraw or @Kijewski can you publish this as `iana-time-zone-haiku` 0.1.1?

Kijewski:
I'm not owner of iana-time-zone-haiku, only iana-time-zone. But I tested the current HEAD on the actual system (well, in a VM), and it works fine.

astraw:
@Kijewski thanks. I just sent you an oversight to be publisher of iana-time-zone-haiku and also published it.

Issue #72, opened 2022-10-08
CI builds are broken on macOS with rust 1.38


astraw:
Our Github Actions CI builds are broken on macOS with rust 1.38.

This doesn't seem to be a bug with `iana-time-zone` itself, rather with some other component. For example, the macos 1.38 build for commit 7baa35e075fea0cb57c28399398978c78e685ad5 succeeded [with runner 20220917.1](https://github.com/strawlab/iana-time-zone/actions/runs/3157145385/jobs/5137664509) but failed [with runner 20220929.2](https://github.com/strawlab/iana-time-zone/actions/runs/3201189235/jobs/5228909128).

I guess the cause is that the Github runner has switched to Apple clang 14.0.0, which seems to have some incompatibility with older versions of rust. The previous version of clang used on the Github runner was 13.1.6. (See the [new runner](https://github.com/actions/runner-images/blob/macOS-12/20220929.2/images/macos/macos-12-Readme.md) versus the [old runner](https://github.com/actions/runner-images/blob/macOS-12/20220917.1/images/macos/macos-12-Readme.md)).

For the record, the error is:

```
error: linking with `cc` failed: exit code: 1
[8](https://github.com/strawlab/iana-time-zone/actions/runs/3201189235/jobs/5228909128#step:5:9)
  |
[9](https://github.com/strawlab/iana-time-zone/actions/runs/3201189235/jobs/5228909128#step:5:10)
[...]

PR #78, opened 2022-10-14
Scope 'push' GitHub Actions workflow trigger to 'main' branch

lopopolo:
The previous configuration causes all jobs to run twice on every PR. We should consider protecting the main branch and requiring PRs if that is not already the case.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -2,7 +2,8 @@ name: build
 
 on:
   push:
-    branches: [ '**' ]
+    branches:
+      - main
   pull_request:
     branches: [ '**' ]
   schedule:


astraw (approved):

astraw:
The CI failures on Android look like a nightly rust problem as they are not happening on stable nor on nightly on other platforms. The Mac issue is described at #72 and is also unrelated.

I don't understand as much about the github CI admin config as I could, but I'll take your word for it that this is a good thing.

In a similar vein, I just did as you suggested and created a new rule that a PR is required before merging to `main` and a require also minimum of 1 approval.

Relatedly, I made you and @Kijewski both admins on this repo, so we now mostly have a bus factor of 3. Also at the moment I think only @Kijewski can publish releases to crates.io. Happy to also add you there if you like, too.

lopopolo:
Thanks for making those changes @astraw. I think I'd be more comfortable without publisher privileges for now if that's ok. If there's a need to add another publisher later, we can revisit that though.

lopopolo:
@astraw I believe https://github.com/rust-lang/cc-rs/pull/728 will fix the issues with android on nightly since the latest nightly made changes to which Android NDK is used: https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/1.2E65.20and.20the.20Android.20NDK

PR #85, opened 2022-10-28
Use less cross-compilations in CI tests

### LINKED ISSUE #80: android cross builds are failing on CI with nightly rust
Android cross builds are failing on CI with nightly rust. This started between [build 351](https://github.com/strawlab/iana-time-zone/actions/runs/3214907058) and [build 352](https://github.com/strawlab/iana-time-zone/actions/runs/3220695414) (the macOS failures were another unrelated issue #72 which has now been solved).

The key error from the CI logs seems to be:

```
aarch64-linux-android/bin/ld: cannot find -lunwind
```

Kijewski:
We aren't using any assembly code, but only the official C interface for our Android and MacOS implementation. It's exceedingly unlikely that the compilation for e.g. AArch64 could fail if it succeeds for x86-64.

Should hopefully resolve #80.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -119,9 +119,6 @@ jobs:
           - x86_64-unknown-illumos
           - x86_64-unknown-netbsd
           - x86_64-linux-android
-          - i686-linux-android
-          - arm-linux-androideabi
-          - aarch64-linux-android
           - sparcv9-sun-solaris
         versions:
           - ""
@@ -165,8 +162,6 @@ jobs:
           - "1.48"
           - stable
         target:
-          - aarch64-apple-ios-sim
-          - aarch64-apple-ios
           - x86_64-apple-ios
         versions:
           - ""


astraw (commented):
I agree we don't need to test all the architectures, but why remove `sparcv9-sun-solaris`?

.github/workflows/rust.yml
```diff
@@ -119,10 +119,6 @@ jobs:
           - x86_64-unknown-illumos
           - x86_64-unknown-netbsd
           - x86_64-linux-android
-          - i686-linux-android
-          - arm-linux-androideabi
-          - aarch64-linux-android
-          - sparcv9-sun-solaris
```
astraw, line 125: Why drop solaris? It seems to me like we should keep that.

astraw (commented):
I think we can merge this after addressing the solaris issue.

Kijewski:
The link-time error for android was fixed in cross-rs, but the fix was not released, yet: https://github.com/cross-rs/cross/issues/1222#issuecomment-1474110816

.github/workflows/rust.yml
Kijewski: You're right. I don't know why I dropped the solaris line. :)

Kijewski:
I guess we only need to replace two lines in the changelog and we do a release:

```diff
-## [0.1.54] - unreleased
+## [0.1.54] - 2023-03-21
 ### Changed
-- replace `winapi` dependency with `windows-sys` ([#91](https://github.com/strawlab/iana-time-zone/pull/91))
+- replace `winapi` dependency with `windows` ([#97](https://github.com/strawlab/iana-time-zone/pull/97))
```

I you want I can do that tomorrow.

astraw:
Yes, it would be great if you do a release.

PR #86, opened 2022-11-02
Update `cxx-build` version constraint in haiku crate

lopopolo:
cxx-build before v1.0.79 did not properly declare its minimum version bound on `codespan-reporting`, which caused `cxx-build` (and `iana-time-zone-haiku`) to fail building when used with a `-Zminimal-versions` lockfile.

I previously reported this upstream to `cxx`:

- https://github.com/dtolnay/cxx/pull/1111

The fix was included in v1.0.79.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -302,8 +302,10 @@ jobs:
         runs-on: [ubuntu-20.04, windows-2022, macos-12]
         toolchain:
           - stable
+          - nightly
         versions:
           - ""
+          - "-Zminimal-versions"
     runs-on: ${{ matrix.runs-on }}
     env:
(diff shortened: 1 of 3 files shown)


lopopolo:
Ahh I thought the commit in codespan-reporting – https://github.com/brendanzab/codespan/issues/322#issuecomment-1299621536 – had been released on crates.io, but that turns out not to be the case. I bumped upstream to see if they can release things, and then I'll have to bump in cxx again.

lopopolo:
Blocked on https://github.com/brendanzab/codespan/issues/348.

lopopolo:
This PR is obsolete now that #106 is merged and `cxx` is no longer used in the haiku platform support crate.

PR #91, opened 2022-12-13
Use windows-sys

Kijewski:


astraw (approved):
I checked that this runs on my Windows machine and looked through it for a few minutes. I don't see any obvious problems. Although I'm certainly no expert in the relevant APIs, I will approve this now but would welcome another look by any interested party.

astraw:
Thanks. Sorry this took a while to review.

djc:
So chrono 0.4 is still the active release branch, and is currently still maintaining an MSRV of 1.38. We're working on 0.5 (which requires 1.48) but that's not very close to being done. Would it be a big ask to request that this project keeps its MSRV at 1.38 for a bit longer?

PR #96, opened 2023-01-23
WIP: Windows, manually

Kijewski:
This is incredibly ugly, but removes the windows-sys dependency altogether.

Since none of the Vtables for {IUnknown, IInspectable, ITimeZoneOnCalendar} where exported by windows-sys 0.42 anyway, half of the code already was part the our source.

Tested successfully with `x86_64-pc-windows-msvc` and `i686-pc-windows-msvc`. It compiles successfully for `aarch64-pc-windows-msvc`, but I have no device to test it on.


Kijewski:
Tested successfully with {x86_64,i686}-pc-windows-{msvc,gnu}. It cross-compiles successfully for aarch64-pc-windows-msvc, but I have no device to test it on.

src/windows/hstring.rs
```diff
@@ -0,0 +1,102 @@
...
+macro_rules! w {
+    ($s:literal) => {{
+        const INPUT: &[u8] = $s.as_bytes();
+        const OUTPUT_LEN: usize = $s.len() + 1;
+        const OUTPUT: [u16; OUTPUT_LEN] = {
+            let mut buffer = [0; OUTPUT_LEN];
+            let mut pos = 0;
+            while pos < INPUT.len() {
+                // ensure that the input string is ASCII and does not contain any NUL characters
+                let _ascii_only = [0][if matches!(INPUT[pos], 1..=0x7f) { 0 } else { 1 }];
```
astraw, line 57: The ASCII check isn't used for anything... but presumably it should be?

astraw:
I confirm this is working for me on `x86_64-pc-windows-msvc`. That said, this gets really deep into lots of `unsafe` and APIs I'm not very familiar with. Thus, I wrote a request for guidance at https://github.com/microsoft/windows-rs/issues/2317 which I hope gets some feedback.

src/windows/hstring.rs
Kijewski: Actually, no. I removed the UTF-8 → UCS2 conversion from the source, and the line is meant to throw an error at compile time if not all characters where in ASCII range. [Panicking in `const` contexts](https://caniuse.rs/features/const_panic) is only stable since 1.57, so I used the older workaround to access an array out of bounds, which is caught by the compiler. I only gave the result a name, so it would occur in the error message quoting the offending line

Kijewski:
Thank you for opening the thread over at windows-rs!

I added another commit to make the error checking a little easier to read. I think the code is not *too* bad. It's just not rusty, and more like C++ written in Rust. :-/

astraw:
I think this is superseded by #97. I will close this for now. (I'll be glad if we don't have to maintain this amount of `unsafe`.) Thanks for your continued efforts!

PR #97, opened 2023-01-30
use windows crate directly

astraw:
This drops a lot of `unsafe` code on Windows. For now, this pulls in the `windows` crate using git. When this is released it we can update this PR.

See https://github.com/microsoft/windows-rs/issues/2317 and https://github.com/microsoft/windows-rs/pull/2318

Thanks to @Kijewski @kennykerr and @riverar


riverar:
Sweet! ✌️

astraw:
(The failing tests are unrelated noise - they were already failing on a scheduled rebuild of the main branch. I guess it has something to do with the recently released rust 1.67, cross compilation, and freebsd not playing nicely together.)

src/tz_windows.rs
```diff
@@ -1,312 +1,13 @@
...
-};
-
-const CO_E_NOTINITIALIZED: HRESULT = -2147221008;
-
-impl From<HRESULT> for crate::GetTimezoneError {
-    fn from(orig: HRESULT) -> Self {
-        std::io::Error::from_raw_os_error(orig).into()
+impl From<windows::core::Error> for crate::GetTimezoneError {
+    fn from(_orig: windows::core::Error) -> Self {
+        crate::GetTimezoneError::OsError
```
Kijewski, line 5: This line could be made ↓ so the error message does not get lost:

```rust
GetTimezoneError::IoError(io::Error::new(io::ErrorKind::Other, orig))
```

astraw: Nice. I put this in 1b898d1237192b54e831ba56e7e0b88152480e97.

Kijewski:
Windows-rs 0.46 was released, and should be usable for us now.  :tada:

lopopolo (approved):
Looks good and I love dropping all of this unsafe. Thanks all!

kennykerr:
:shipit:

astraw:
> Looks like `windows 0.47` already dropped: can we get a bump and release to that?

Done in v0.1.55

MarijnS95:
That's fast, thanks!

MarijnS95:
> Looks like `windows 0.47` already dropped: can we get a bump and release to that?
> 
> @Jake-Shadle this is also a prime candidate for `windows-bindgen` except for the use of COM objects, for which we're waiting on @kennykerr to separate manual code out into `windows-core` first; and then hope that crate releases less often if breaking changes are only in the metadata and generated output.

I recently tried this now that `windows-bindgen` is capable of generating "minimal" bindings that leverage `windows-core 0.51` rather than the full `windows` crate.

Unfortunately/obviously this doesn't suffice the MSRV 1.48  policy of this crate thanks to `edition = "2021"` in the `windows-core` crate.

Diff is here: https://github.com/MarijnS95/iana-time-zone/compare/windows-core

I generated the bindings without `--config package` because this crate really shouldn't have Windows features leaking into its `[features]` array (beyond those being appended at the end of `Cargo.toml`). Unfortunately this means the [_whole_ `Windows.Globalization.Calendar` type](https://microsoft.github.io/windows-docs-rs/doc/windows/Globalization/struct.Calendar.html) is generated including function signatures that _depend on other types/modules_, **without `cfg` guards**. All these extra types (and all their dependencies, etc) have to be included to make the `Windows` module compile; while users may end up not having to download the `windows` crate as a result, they may end up compiling more code that way.

Still appears to be a half-second (approx!) win though, running:

```sh
cargo clean && cargo b --timing
```

Before:
afbeelding

After:
afbeelding

CC @kennykerr, you might be interested in this. Note also that the compiler suggested that `Windows.Foundation.DateTime` and `Windows.Foundation.TimeSpan` could have been imported from `windows_core::imp::*`, should/could `bindgen` take care of that?

PR #103, opened 2023-05-07
CI: stop using actions-rs/toolchain

astraw:
This fixes the deprecation warning with the Github action `actions-rs/toolchain`. See https://github.com/actions-rs/toolchain/issues/221.


Kijewski (approved):
Ah, without any external dependency. That's great!

Issue #89, opened 2022-12-12
Feature Flag Haiku OS Support To Remove C++ Dependency If Not Used By Library Consumer 


LuisOsta:
Some time ago `iana-time-zone` [added support](https://github.com/strawlab/iana-time-zone/tree/main/haiku) for HaikuOS. Which ended up requiring a dependency on C++ in order to work. Which means that all end consumers of `iana-time-zone` are also dependent on C++.

Would it be possible to feature gate the Haiku OS support that only consumers that actually need to support Haiku OS end up with the C++ dependency? I am dependent on `iana-time-zone` through the `chrono`[ cargo package](https://github.com/chronotope/chrono) and would love to be able to remove that C++ dependency but its not possible while continuing to use `chrono`.

For context we are currently working on a Self Sovereign Identity library in Rust and are evaluating options for how to move forward - https://github.com/knox-networks/ssi/issues/71. Would love to continue using `chrono` and `iana-time-zone` but let me know if feature flagging isn't something you can do.

Kijewski:
Please see #88. The build script is not being executed unless the target platform is actually haiku, so even if you don't have a c++ compiler installed, it should still work for all other platforms.

LuisOsta:
Yeah the build scripts are not being executed but the dependency to the haiku crate and hence C++ still remains. Would it not be possible to remove that dependency via feature flag?

Here's the relevant sample of the Cargo.lock:
```
[[package]]
name = "chrono"
version = "0.4.22"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "bfd4d1b31faaa3a89d7934dbded3111da0d2ef28e3ebccdb4f0179f5929d1ef1"
dependencies = [
 "iana-time-zone",
 "js-sys",
 "num-integer",
 "num-traits",
 "serde",
 "time 0.1.43",
 "wasm-bindgen",
 "winapi",
]
...

[[package]]
name = "iana-time-zone"
version = "0.1.51"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "f5a6ef98976b22b3b7f2f3a806f858cb862044cfa66805aa3ad84cb3d3b785ed"
dependencies = [
 "android_system_properties",
 "core-foundation-sys",
 "iana-time-zone-haiku",
 "js-sys",
 "wasm-bindgen",
 "winapi",
]

...
[[package]]
name = "iana-time-zone-haiku"
version = "0.1.1"
[...]

Issue #88, opened 2022-11-16
Feature flag platforms


Sytten:
Currently all platforms are enabled by default and pulling their dependencies.
I recently tried upgrading to 1.0.53 and it added a bunch of things for haiku that I really don't care about since that's not a target for our software.
It would be nice to feature gate each platform.

astraw:
This shouldn't be happening. Dependencies are pulling in only for the target being built. Can you provide steps to reproduce?

Kijewski:
It's the same problem as in #61. Dependencies that are only needed in other platforms are still downloaded and evaluated, but they don't get compiled. That is a shortcoming of cargo that we can do nothing about.

You have to file an issue with [rust-lang/rust](https://github.com/rust-lang/cargo) if you want to see this problem fixed, but I daresay it's already known.

Sytten:
Make sense @Kijewski. There is one exception though.
If you put the dependencies as optional and enable them with a feature flag (and default features to all platforms).

Kijewski:
But that would break all the usages, that only indirectly depend on iana-time-zone, e.g. through chrono. IMO this bug has to be fixed in cargo itself, and working around the problem downstream does only hurts usability.

Sytten:
Only if you don't add those feature flags in the default features no?

Kijewski:
By default it should work everywhere, for every target platform, but maybe the number of supported platforms could be restricted by using  `no-default-features = true`.

But unless I'm mistaken, the wasm support pulls in far more dependencies than the haiku support.

astraw:
Thanks all for the discussion. I am closing this for now.

lopopolo:
hi @Sytten, following up here, we significantly reduced the impact of `iana-time-zone-haiku` on lockfile weight with this PR: https://github.com/strawlab/iana-time-zone/pull/106.

Let us know if this resolves your concerns!

Sytten:
Ha thanks! I will give it is a shot. cc is acceptable as a dependency.

PR #104, opened 2023-05-10
Allow crate users to opt out of some platform support

lopopolo:
We have gotten several inbound tickets with users expressing concern over the inclusion of `cxx` to support the haiku platform. Additionally, for my personal use, I would prefer to be able to prune things like `wasm-bindgen` from my lockfile when not expressly enabling Wasm platform support, similar to the approach the `getrandom` crate takes.

This commit adds cargo features which allow disabling platform support for each platform that requires a third party dependency. Currently, that set of platforms is:

- android
- all apple targets
- haiku
- wasm
- windows

This commit adds a new default feature, `platform-all`, which retains the existing behavior of all platforms being enabled by default. `platform-all` is a meta feature which enables all of the other platform support features.

For example, to disable haiku support, users would add the following to their `Cargo.toml`:

```toml
iana-time-zone = { version = "0.2", default-features = false, features = ["platform-android", "platform-apple", "platform-wasm", "platform-windows"] }
```

The addition of a new default feature is a breaking change because it may break the build for folks currently depending on this crate with `default-features = false`.

I'm not sure if this is the direction we want to go, but it would address the ask from these tickets:

- https://github.com/strawlab/iana-time-zone/issues/89
- https://github.com/strawlab/iana-time-zone/issues/88

I know that cargo won't build deps unless compiling for a particular platform, but practically, most users don't want or need Haiku or Wasm support. These targets have heavy dependencies which end up in the lock file. Deps in the lock file have a carrying cost in terms of supply chain risk and need for patching/updating.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "iana-time-zone"
 description = "get the IANA time zone for the current system"
-version = "0.1.56"
+version = "0.2.0"
 authors = [
     "Andrew Straw <strawman@astraw.com>",
     "René Kijewski <rene.kijewski@fu-berlin.de>",
@@ -15,27 +15,44 @@ readme = "README.md"
 edition = "2018"
 
 [features]
+default = ["platform-all"]
+# Allow users to opt out of platform support for platforms that require third
+# party dependencies. Disabling any of these platforms (i.e. with
+# `default-features = false`) will prevent the platform dependencies from
+# appearing in downstream applications' `Cargo.lock`.
+platform-all = [
+  "platform-android",
+  "platform-apple",
+  "platform-haiku",
+  "platform-wasm",
+  "platform-windows",
+]
+platform-android = ["android_system_properties"]
+platform-apple = ["core-foundation-sys"]
+platform-haiku = ["iana-time-zone-haiku"]
+platform-wasm = ["js-sys", "wasm-bindgen"]
+platform-windows = ["windows"]
 # When enabled, the library will succeed to compile for unknown target platforms, and return an `Err(GetTimezoneError::OsError)` at runtime.
 fallback = []
 
 [target.'cfg(target_os = "android")'.dependencies]
-android_system_properties = "0.1.5"
+android_system_properties = { version = "0.1.5", optional = true }
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
-core-foundation-sys = "0.8.3"
+core-foundation-sys = { version = "0.8.3", optional = true }
 
 [target.'cfg(target_os = "windows")'.dependencies]
-windows = { version = "0.48.0", features = [ "Globalization" ] }
+windows = { version = "0.48.0", features = [ "Globalization" ], optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
-js-sys = "0.3.50"
-wasm-bindgen = "0.2.70"
+js-sys = { version = "0.3.50", optional = true }
+wasm-bindgen = { version = "0.2.70", optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dev-dependencies]
 wasm-bindgen-test = "0.3"
 
 [target.'cfg(target_os = "haiku")'.dependencies]
-iana-time-zone-haiku = { version = "0.1.1", path = "haiku" }
+iana-time-zone-haiku = { version = "0.1.1", path = "haiku", optional = true }
 
 [workspace]
 members = [".", "haiku"]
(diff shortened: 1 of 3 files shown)


lopopolo:
it looks like the android build fails the same way as in #105, so I don't think I've broken anything with the new Cargo features.

astraw:
I'm not sure this merits a semver breaking change - the API has not changed. Also, the build should not break for 99.9% of people, either. Specifically, `chrono` does not disable the default features in its `Cargo.toml`.

That said, I'm undecided about this overall and am still thinking about it. @Kijewski @djc I would be interested in your thoughts.

djc:
> I'm not sure this merits a semver breaking change - the API has not changed.

Hmm -- the API surely has changed for people who depend on iana-time-zone with `default-features = false`, right?

> Additionally, for my personal use, I would prefer to be able to prune things like `wasm-bindgen` from my lockfile when not expressly enabling Wasm platform support, similar to the approach the `getrandom` crate takes.

Why? What are the downsides of having more crates in your `Cargo.lock` file? (I'm not trying to suggest there aren't any, but I think it's valuable for this PR to enumerate what the downsides are specifically.) The reporter for #89 never answered this question. The reporter in #88 wrote:

> I am just a bit paranoid with my lockfile :)

These don't seem like strong motivations to me, and a lot of the "complaints" seem to be from people who don't really understand how Cargo works in this case.

FWIW, I wrote up a more general solution in [a Cargo issue](https://github.com/rust-lang/cargo/issues/11313), but have so far not found the time to go back and turn this into an RFC with the level of detail that it needs.

Ultimately this solution is kinda neat and doesn't impose a lot of complexity, so I'm not against it. It still adds some complexity and doing a semver-breaking release for it might cause a little pain over time due to duplicate dependencies? (Are there other API-breaking changes that you'd like to make but haven't so far?)

Kijewski:
The main problem I see: How would one use the feature? E.g. how would `chrono` use the this library?

```toml
[dependencies]
iana-time-zone = { default-features = false }

[features]
default = ["platform-all"]
platform-all = ["iana-time-zone/platform-all"]
platform-amiga = ["iana-time-zone/platform-amiga"]
platform-irix = ["iana-time-zone/platform-irix"]
platform-pcdos = ["iana-time-zone/platform-pcdos"]
```

How would a user of `chrono` use this feature?

```toml
[dependencies]
chrono = { default-features = false }

[features]
default = ["platform-all"]
platform-all = ["chrono/platform-all"]
platform-amiga = ["chrono/platform-amiga"]
platform-irix = ["chrono/platform-irix"]
platform-pcdos = ["chrono/platform-pcdos"]
```

Should there be an avalanche of updates if we add a new platform?

IMO <https://github.com/rust-lang/cargo/issues/11313> is the only feasible solution for this problem.

The breaking change would actually not be a problem, AFAIS. Dtolnay's [semver-trick](https://github.com/dtolnay/semver-trick) would work fine. A last version in 0.1 would re-export 0.2 with its default features activated.

lopopolo:
`chrono` is not the only way to use this library. It wouldn't be necessary for `chrono` to expose these features if there would be too much complexity.

One option though would be for `chrono` to only enable features for tier 1 platforms by default. If applications require support for more exotic platforms, they can depend on `iana-time-zone` directly and activate more of its features. Because features are additive and unified in the entire dep graph, `chrono`'s use of `iana-time-zone` would inherit full platform support.

Kijewski:
Chrono was only an example.

Maybe features `tier-1` to `tier-3` would be easier to use than a feature per platform? With `default  = ["tier-1", "tier-2"]`? And the multitude of haiku users would have to depend on this library explicitly.

djc:
I don't think it makes sense for chrono to disable platform support for some platforms by default on this basis.

lopopolo:
As a shorter term fix while we think about the design implications here, I've put up a PR I hope we can release as `iana-time-zone-haiku` v0.1.2 which replaces `cxx` and `cxx-build` with a C ABI-style binding and `cc`:

- https://github.com/strawlab/iana-time-zone/pull/106

djc:
> The rust API is unchanged here. So the semver question is whether requiring downstream to fiddle with `Cargo.toml` counts as a breaking change. To me, this seems related to the question of whether MSRV bumps should be considered a breaking semver change, to which I think the consensus is no, although debatable.

If functionality in this crate regresses while the feature flags are unchanged, I would consider that a semver violation. If iana-time-zone previously provided support for all platforms with the default features, then changing the default features to only support a subset of platforms is not a semver-compatible change.

I think the energy spent on this would be better directed at a solution at the Cargo level.

astraw:
Yes, I'm inclined to suggest that with #106 the short term pain is minimized and that we should close this issue and direct future energy at the Cargo level. I am closing this now but please re-open if you think we should indeed discuss further.

Issue #108, opened 2023-06-06
missing openwrt timezone support


cppcoffee:
Rust example code:

```rust
fn main() -> anyhow::Result<()> {
    let tz_str = iana_time_zone::get_timezone()?;
    println!("The current time zone is: {}", tz_str);
    Ok(())
}
```

On OpenWrt output:

```
Error: No such file or directory (os error 2)

Caused by:
    No such file or directory (os error 2)
```

Openwrt timezone config:

```
$ cat /etc/config/system
config system
        option timezone 'CST-8'
        option zonename 'Asia/Shanghai'
...
```

Expected openwrt output +08:00

Associate issue: https://github.com/chronotope/chrono/issues/1129

Is openwrt timezone support considered?

cppcoffee:
I temporarily executed `echo "Asia/Shanghai" > /etc/timezone` and the output was fine.

Is this approach recommended?

Kijewski:
We should implement implement the support for OpenWRT ourself, and parse `/etc/config/system` accordingly. I'd have to rely on you to test the PR, though. :)

Can you verify that this is the target platform: [mips64-openwrt-linux-musl](https://doc.rust-lang.org/nightly/rustc/platform-support/mips64-openwrt-linux-musl.html)?

cppcoffee:
Executing `echo "Asia/Shanghai" > /etc/timezone`, chrono Local outputs the time utc.

```rust
use chrono::{Local, Utc};

fn main() -> anyhow::Result<()> {
    let tz_str = iana_time_zone::get_timezone()?;
    println!("The current time zone is: {}", tz_str);
    println!("{}", Local::now());
}
```

Output:
```
The current time zone is: Asia/Shanghai
Local: 2023-06-06 08:19:39.322573799 +00:00
Utc: 2023-06-06 08:19:39.322880350 UTC
```

https://github.com/strawlab/iana-time-zone/issues/108#issuecomment-1578123510

Is the chrono crate issue?

cppcoffee:
No problem, leave it to me to do the testing. The platform we are using is mipsel-unknown-linux-uclibc from the https://doc.rust-lang.org/nightly/rustc/platform-support.html list.

We have compiled the rust docker for mipsel-ubknow-linux-uclibc at https://hub.docker.com/repository/docker/cppcoffee/rust_uclibc_openssl/tags?page=1& ordering=last_updated

cppcoffee:
[...]

PR #109, opened 2023-06-07
Implement for OpenWRT

Kijewski:
Closes issue #108.


Kijewski:
So, OpenWRT is just a musl Linux. This means we cannot tell at compile time if we need to add support for OpenWRT or not. Because big, normal-desktop distros like Alpine Linux use musl, too, we cannot opt-out of the normal for musl targets, either.

cppcoffee:
I have written a demo that outputs the time zone and it is correct.

Cargo.toml

```
[dependencies]
anyhow = "1.0.71"
chrono = "0.4.26"
iana-time-zone = { git = "https://github.com/Kijewski/iana-time-zone.git", branch = "pr-openwrt" }

[patch.crates-io]
iana-time-zone = { git = "https://github.com/Kijewski/iana-time-zone.git", branch = "pr-openwrt" }
```

src/main.rs

```
use chrono::{Local, Utc};

fn main() -> anyhow::Result<()> {
    let tz_str = iana_time_zone::get_timezone()?;
    println!("The current time zone is: {}", tz_str);
    println!("Chrono Local: {}", Local::now());
    println!("Chrono Utc: {}", Utc::now());
    Ok(())
}
```

output:

```
# ./demo-uclibc
The current time zone is: Asia/Shanghai
Chrono Local: 2023-06-07 06:20:29.643589639 +00:00
Chrono Utc: 2023-06-07 06:20:29.643894306 UTC
```

It looks like timezone is right.

Kijewski:
Thank you for testing the PR!

astraw (approved):

astraw:
Thanks, both!

astraw:
Released as 0.1.57

cppcoffee:
Great job!

cppcoffee:
Awesome!

cppcoffee:
Excellent!

PR #110, opened 2023-08-09
bump haiku dependency

extrawurst:
This bump is needed to get rid of `syn1` in the dependency tree when using `iana-time-zone`.

After this being merged we would need a release


Kijewski:
`cargo update` will automatically use the newer `iana-time-zone-haiku` version, so this should not be needed.

If you are manually maintaining your `Cargo.lock`, then `cargo update -p iana-time-zone-haiku --precise 0.1.2` should work for you.

extrawurst:
totally right, thanks for pointing that out

PR #112, opened 2023-08-15
Allow version range for `windows` dependency

Kijewski:
`iana-time-zone` works with any `windows` version since `0.30.0`. We don't have to care which version the user actually uses. We only need to make sure that the selected version is not too old. Let `cargo` decide everything else.

The newest version of `windows`, `0.51.0`, has an msrv of `1.56`. If a user of `iana-time-zone` is not able or willing to update their rust installation, and want to continue using rustc `1.48`, they can select an older version of the `windows` crate, and still receive any updates `iana-time-zone` might have in the future.

Resolves #111.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.1.58] - unreleased
+### Changed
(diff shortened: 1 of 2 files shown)


MarijnS95:
FYI if the MSRV range wasn't this strict this crate could migrate to `windows-core` and a "minimal" set of bindings emitted with `windows-bindgen`, more details here: https://github.com/strawlab/iana-time-zone/pull/97#issuecomment-1684444549

Maybe something for the future 👍

astraw:
This is redundant now with other recent merges. Closing, but please re-open (and update) if this is in error.

PR #113, opened 2023-08-15
CI: use `cargo ndk` to build android versions

Kijewski:


Kijewski:
There seems to be some problem in cargo-ndk, tt always wants to build for `armeabi-v7a` and `arm64-v8a`. But I guess that's OK. That's a 32bit and 64bit platform, and compiling for `i686` and `x86_64` would not give us any more information, since we aren't running the tests anyway. We only compile them.

astraw:
Fantastic! Thanks so much.

Issue #115, opened 2023-08-18
[discuss] increase MSRV


lopopolo:
Ref: https://github.com/strawlab/iana-time-zone/pull/97#issuecomment-1684444549

With Debian Bookworm becoming the latest stable Debian release in June, can we consider raising this crate's MSRV? Bookworm ships `rustc` 1.63.0.

lopopolo:
Apologies for not cc'ing folks (specifically the chrono team); I'm on a tiny screen.

Kijewski:
Chrono is already edition=2016 and msrv=1.56: <https://docs.rs/crate/chrono/0.4.26/source/Cargo.toml.orig>. So if we upgrade our requirements, we wouldn't not break anything for them. 

Since syn upgraded to edition=2021, I think almost everyone had to update to rust 1.56.

astraw:
We just updated the MSRV to 1.61 in #157

PR #117, opened 2023-08-20
Use `windows-core` with embedded bindings via `windows-bindgen`

### LINKED ISSUE #116: Implement windows support using `windows-bindgen`
> > Looks like `windows 0.47` already dropped: can we get a bump and release to that?
> >
> > @Jake-Shadle this is also a prime candidate for `windows-bindgen` except for the use of COM objects, for which we're waiting on @kennykerr to separate manual code out into `windows-core` first; and then hope that crate releases less often if breaking changes are only in the metadata and generated output.
>
> I recently tried this now that `windows-bindgen` is capable of generating "minimal" bindings that leverage `windows-core 0.51` rather than the full `windows` crate.
>
> Unfortunately/obviously this doesn't suffice the MSRV 1.48  policy of this crate thanks to `edition = "2021"` in the `windows-core` crate.
>
> Diff is here: https://github.com/MarijnS95/iana-time-zone/compare/windows-core
>
> I generated the bindings without `--config package` because this crate really shouldn't have Windows features leaking into its `[features]` array (beyond those being > appended at the end of `Cargo.toml`). Unfortunately this means the [_whole_ `Windows.Globalization.Calendar` type](https://microsoft.github.io/windows-docs-rs/doc/windows/Globalization/struct.Calendar.html) is generated including function signatures that _depend on other types/modules_, **without `cfg` guards**. All these extra types (and all their dependencies, etc) have to be included to make the `Windows` module compile; while users may end up not having to download the `windows` crate as a result, they may end up compiling more code that way.
>
> Still appears to be a half-second (approx!) win though, running:
>
> ```sh
> cargo clean && cargo b --timing
> ```
>
> Before:
> afbeelding
>
> After:
> afbeelding
>
> CC @kennykerr, you might be interested in this. Note also that the compiler suggested that `Windows.Foundation.DateTime` and `Windows.Foundation.TimeSpan` could have been imported from `windows_core::imp::*`, should/could `bindgen` take care of that?

_Originally posted by @MarijnS95 in https://github.com/strawlab/iana-time-zone/issues/97#issuecomment-1684444549_

MarijnS95:
Closes #116

As discussed in https://github.com/strawlab/iana-time-zone/pull/97#issuecomment-1490030417 `iana-time-zone` might not have to pull in the whole `windows` crate, but only generate the bindings (one single type and function call) that it needs and use simplified/stripped-down windows components from the (much lighter!) `windows-core` crate.

As also brought up in https://github.com/strawlab/iana-time-zone/pull/97#issuecomment-1684444549 however, this approach currently has as few drawbacks:

- No `cfg` around APIs. The whole `Windows.Globalization.Calendar` type is generated, and a few functions require external types, which also require more types ,leading to a bit of (compile) overhead (and generated file bloat) that wasn't there with the `windows` crate;
- MSRV bump to 1.56 (#115).

On my machine this appears to be about half a second faster to compile.

---

Opening this as draft to put some visibility to the changes I've already made, and to have a place to discuss how/where we could improve this; allowing it to be merged (if folks are happy with this) as soon as there is an applicable MSRV bump.


Kijewski:
Thank you for working on this! I propose this new CI test (.github/workflow/rust.yml):

```yaml
  test-windows-api-gen:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout
        uses: actions/checkout@v3
      - name: Install Rust
        run: |
          rustup toolchain install stable --profile minimal --no-self-update
      - run: cargo run --package api_gen
      - run: git diff --exit-code
```

This way we know if anything has changed upstream, and users can be sure that the code is what it claims to be. :)

Kijewski:
The generated code works with `windows-core==0.50.0`, too. This version has a msrv of 1.48, so this PR would be an advantage over the old version without any downsides.

I would allow a version range as I proposed in #112: `windows-core = { version = ">= 0.50, < 0.52" }`. We already have tests in place if compiling with `-Zminimal-versions` works.

MarijnS95:
> Thank you for working on this! I propose this new CI test (.github/workflow/rust.yml):

I hadn't yet pushed this from my other `windows-bindgen` crates, but have done so now. I prefer `git status --porcelain` because it includes untracked files, in case `api_gen` starts spitting out more files in future updates/changes but the submitter forgets to check them in.

Not sure what we need most `rustup` calls for: GitHub's virtual environments already come preinstalled with all that.

> The generated code works with `windows-core==0.50.0`, too. This version has an msrv of 1.48, so this PR would be an improvement without any downsides.
> I would allow a version range as I proposed in #112: `windows-core = { version = ">= 0.50, < 0.52" }`.

Nice catch! I hadn't even thought of trying this because [`windows-core 0.50` did not have a matching/compatible `windows-bindgen` release](https://github.com/microsoft/windows-rs/pull/2529#issuecomment-1570338796), but for this crate the code generated by `windows-bindgen 0.51` (now that we finally have a release...) appears to be compatible with `windows-core 0.50` (that isn't the case for my other Windows projects though...) :tada:!

There's a strict downside as users of MSRV 1.48 will now have to make sure to _lock_ `windows-core` at `0.50`, this crate will no longer compile out of the box for them. In that sense:

> We already have tests in place if compiling with `-Zminimal-versions` works.

This will only apply to the `wasm` tests whose testing matrix currently combines MSRV 1.48 _with_ `-Zminimal-versions`. The regular `test` job matrix checks MSRV 1.48 with _and without_ `-Zminimal-versions`, which should fail to compile (as suggested in the previous paragraph).

MarijnS95:
Let's approve/run the CI to confirm the above?

Kijewski:
Darn, this will change will increase the MSRV to 1.56 for unrelated platforms, too. Cargo fails if it finds Cargo.toml with `editon = "2021"` anywhere in its dependencies, regardless if the dependency is used or it. Cargo downloads any and all non-optional packages, even if the `cfg(target_os = …)` excluded the package from winding up in the resulting compilation.

IMO it's okay to bump the MSRV. Versions < 1.56 are unusable by now, because fundamental crates like syn are edition = 2021 by now.

MarijnS95:
Aplogies, it was me who forgot to replace `edition = "2021"` with `edition = "2018"` in `api_gen`, but even then its `windows-bindgen` dependency is still `2021`. We can remove that crate from the workspace entirely as it's just a tool to generate the bindings, which are checked in and won't change afterwards. The tool itself isn't intended to be distributed with the crate.

I've removed it so that we can see where we stand now - probably still need to downgrade `windows-core` in the MSRV-1.48 Windows build.

MarijnS95:
Fwiw the `-Zminimal-versions` detailed above might be too harsh, we could even add a `cargo update -p windows-core --precise 0.50.0` to only affect the MSRV of that crate. The rest could remain at latest.

MarijnS95:
@Kijewski where are we at with this?

Kijewski:
Thank you for your work! Sorry that it took so long to get merged.

MarijnS95:
No worries, I just wanted to see the CI run-failure so that I could add something akin to https://github.com/strawlab/iana-time-zone/commit/ed6476b174ec48813b5acc60434a5e930742192a, making sure that Windows+MSRV is only validated for `-Zminimal-versions` (and settling that it's okay that MSRV-compliancy is only there when manually or via `-Zminimal-versions` downgrading some dependencies) :+1:

PR #119, opened 2023-09-19
CI: download pre-built libraries by install action

Kijewski:
This has the advantage that always the newest version is selected without us having to update the version number. (And it looks less like black magic.)


astraw (approved):
looks good to me

PR #126, opened 2023-12-30
Bump version number to 0.1.59

Kijewski:


Kijewski:
Netbsd currently does not cross-compile correctly, but that is unrelated to this PR or #125.

astraw (approved):
It looks good to me. (Although I haven't checked this on Windows as I'm not using that OS much at the moment and this is the only OS with changes. Still, I don't see much potential for problems.)

astraw:
@Kijewski would you like to do the honors of publishing this?

Kijewski:
Sure! Released as 0.1.59. Guten Rutsch in advance! :)

PR #127, opened 2024-02-03
correct windows-core dependency version

astraw:
As noted in https://github.com/strawlab/iana-time-zone/pull/125#issuecomment-1875367853, commit d3a9e0a21f4c4ded969d02ceec0a908cd7cd0ca4 introduced a potentially breaking change. This is the fix.

@MarijnS95 thanks for the tip. Do you know if there a public issue regarding this dependabot bug that we can link to? I spent a little bit of time looking around https://github.com/dependabot/dependabot-core/issues but didn't find anything relevant.

I would release this as 0.1.60.

Shall we then yank 0.1.59?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ android_system_properties = "0.1.5"
 core-foundation-sys = "0.8.3"
 
 [target.'cfg(target_os = "windows")'.dependencies]
-windows-core = { version = ">=0.50, <=0.53" }
+windows-core = { version = ">=0.50, <=0.52" }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 js-sys = "0.3.50"


Kijewski (approved):

Kijewski:
> Shall we then yank 0.1.59?

Yes, I guess this would be a good idea.

MarijnS95 (approved):

MarijnS95:
I'm not aware of an upstream reported issue about this (nor where to report it), but it might be good to include logs.

astraw:
OK, 0.1.60 published and 0.1.59 yanked.

c-git:
> > I think it's usually a bad idea to yank a crate just becaue it has breaking changes
> 
> While you might not be affected this time, other downstream consumers _could_ be affected when e.g. running with `-Zminimal-versions` against a crate version that is broken, selected (indirectly) by certain projects, and not yanked.

Thanks for explaining why it was yanked

PR #130, opened 2024-03-16
Fix: Fix it tries to use wasm_bindgen in wasm32-unknown-emscripten

### LINKED ISSUE #129: panics on wasm32-unknown-emscripten
ssia.
```
thread '<unnamed>' panicked at /home/sevenc7c/.cargo/registry/src/index.crates.io-6f17d22bba15001f/js-sys-0.3.69/src/lib.rs:295:1:
cannot call wasm-bindgen imported functions on non-wasm targets
stack backtrace:
```

sevenc-nanashi:
SSIA. 
closes #129.


astraw:
Thank you.

Can you write a CI test which this fixes? Also, what does "SSIA" mean?

Don't worry about the existing CI test failures. They are unrelated.

Kijewski:
> Also, what does "SSIA" mean?

"Subject Says It All"

> Can you write a CI test which this fixes?

I don't think this is needed. We could only test that compiling for an emscripten target successfully fails at static link time instead of dynamic link time, i.e. during the build instead of only when you run it.

I know too little about emscripten. I guess you can import native JavaScript methods somehow, and get our library working on this target, too. But I guess that's for another PR.

Kijewski (approved):

curiousdannii:
Emscripten fully supports the libc time functions, or C++ std::chrono. So while that's different from what you're currently doing in Linux (looking up zoneinfo manually) it should be completely reliable.

I'm using Rust's chrono library, and I can no longer get timezones in my Emscripten build. If you've never supported Emscripten then I guess they used to use another method? Still pretty unfortunate that things broke during a non-semver-major update. But that should be posted on their repo instead.

PR #131, opened 2024-03-26
Bump `windows-core` to `0.53-0.54` range

MarijnS95:
Depends on #161 

This includes the necessary binding changes for compatibility, provided by `windows-bindgen 0.53-0.54`.  Generated code by `windows-bindgen` now also expects `TryInto` to be in scope via Rust's `prelude`, which is only the case in Rust edition `2021`.


MarijnS95:
Bumping the edition for having `TryInto` in the `prelude` will break MSRV, let's see if we can find a way around that.

MarijnS95:
@astraw I don't see any strategy as to what the policy is regarding MSRV bumps in https://github.com/strawlab/iana-time-zone?tab=readme-ov-file#minimum-supported-rust-version-policy, and it seems likely that an internal version bump for a newer crate that doesn't provide any useful/fixed functionality otherwise (besides getting rid of duplicate deps downstream) isn't very compelling to break MSRV for. Despite that, what are your thoughts? The [requested version here for `windows 0.53` is Rust `1.62`](https://docs.rs/crate/windows-core/0.53.0/source/Cargo.toml).

Cargo.toml
```diff
@@ -12,7 +12,7 @@ license = "MIT OR Apache-2.0"
 keywords = ["IANA", "time"]
 categories = ["date-and-time", "internationalization", "os"]
 readme = "README.md"
-edition = "2018"
+edition = "2021"
```
lopopolo, line 15: Please back out this change. There's a repo-wide PR for this up here: https://github.com/strawlab/iana-time-zone/pull/161.

If it is necessary to get this PR buildable, please update the base branch of this PR to that of #161.

Cargo.toml
```diff
@@ -26,7 +26,7 @@ android_system_properties = "0.1.5"
 core-foundation-sys = "0.8.6"
 
 [target.'cfg(target_os = "windows")'.dependencies]
-windows-core = { version = ">=0.50, <=0.52" }
+windows-core = { version = ">=0.53, <=0.54" }
```
lopopolo, line 29: are we able to lower the bound on this so we can get the crate building with Rust 1.61.0?

Cargo.toml
MarijnS95: As documented, this change is required to support the new compatibility code generated by newer `windows-bindgen`. We can't back it out, we can however rebase this PR on top of yours as done now.

Cargo.toml
MarijnS95: Let me do some experimentation to see how far we can stretch this bound _without_ bumping the generator and hence requirements for MSRV/edition, and keep separate PRs open when we do end up bumping those.

MarijnS95: [`0.53` already requires MSRV `1.62`](https://crates.io/crates/windows-core/0.53.0), the resulting compatibility issues don't even matter at that point.

lopopolo:
whoops sorry, didn't mean to close this. #161 is merged and when I deleted the `dev/lopopolo-2021-edition` branch, this closed. those changes are in `main` now

lopopolo:
@MarijnS95 Just to confirm my understanding of the current state:

* This PR incrementally bumps `windows-core` to versions 0.53.0 and 0.54.0, which are now significantly behind the latest available release.
* The specified version range (`>=0.53, <=0.54`) requires bumping the MSRV to Rust 1.62.0.
* Supporting a broader range isn’t possible here due to breaking changes in the output of `windows-bindgen`, which is also upgraded as part of this PR.

Looking ahead to your other PR (#133):

* It bumps `windows-bindgen` to v0.59.
* It expands the supported windows-core version range to `>=0.56, <=0.61`, which includes the latest release.
* It maintains an MSRV of Rust 1.62.0 (aligned with `windows-core@0.56`).

Assuming I’ve summarized this correctly, this means that if we’re comfortable raising our MSRV by _one version_ (to Rust 1.62.0), we can adopt a significantly more up-to-date version range of `windows-core`, covering both the newest version and compatibility with our MSRV.

Could you please confirm this assessment?

If this understanding is correct, the next step would be discussing with the other maintainers and the Chrono team whether this MSRV increase is acceptable.

Thanks again for your patience and for driving these improvements. I really appreciate your efforts here!

astraw:
Whoops, now I apparently accidentally deleted the git branch associated this PR. Sorry! (Actually, I thought I was deleting the `dev/lopopolo-2021-edition` branch which was already merged into `main`.) This closed this PR and Github is not letting me simply reopen it. I hope @lopopolo or @MarijnS95 you have the branch locally and push it to either re-open the PR or, worst case, open a new PR.

Many thanks for your ongoing work on this - I really appreciate it. I apologize for this.

@MarijnS95 I also missed your question about the MSRV policy. My main thought is that increases up to chrono's level are likely fine and beyond that should ideally be done in consultation with the chrono team.

lopopolo:
@astraw you can just push up the branch with the same name on current main and it will let you reopen the PR.

I'm used to work GitHub where stacking branches like this would automatically update the base of a PR to the primary branch on merge, but I guess that doesn't work with forks.

astraw:
OK, I was able to restore the branch I deleted and the GitHub let me reopen the PR. Sorry for the noise. (I didn't anticipate that deleting the name of a merged branch would cause this. Now I know how GitHub keeps force push commits associated with the same PR.)

lopopolo:
@astraw the bump we just landed to 1.61.0 matches Chrono. In order to move forward here, we have to bump again to 1.62.0.

MarijnS95:
Looks like GitHub is just broken. It _used to_ automatically reset the target branch to `main` again when a "random" target branch disappears :upside_down_face: 

---

@lopopolo your points above are all correct, expanding on these:

> Supporting a broader range isn’t possible here due to breaking changes in the output of `windows-bindgen`, which is also upgraded as part of this PR.

Correct, both forwards and backwards.

> It expands the supported `windows-core` version range to `>=0.56, <=0.61`, which includes the latest release.

Yes, and I don't think/remember if there was another combination that is useful to us (i.e. if we use `windows-bindgen 0.58`, can we use an even wider version range?). `windows-bindgen 0.59` is a massive improvement over older versions and generates a lot less code.

I've pushed a preliminary MSRV bump to this PR _just to see if the CI is happy with these claims_ (here and in #133), feel free to pull that in through a separate PR when upstream consumers like `chrono` are ready to update.

---

> I didn't anticipate that deleting the name of a merged branch would cause this. Now I know how GitHub keeps force push commits associated with the same PR

Yeah a PR tracks _a branch_, wherever that branch ends up pointing to (and doesn't prune "unreachable" commits, since they're commonly still "reachable" from the UI here, like the `force-pushed from X to Y` event markers).

lopopolo:
It looks like CI is passing with the MSRV bump. Thanks again @MarijnS95 for your efforts!

I’d like to move forward with this proposal, but recognize it might cause some friction for downstream consumers, so I want to ensure we’re fully aligned before proceeding.

Specifically, we’re proposing to raise this crate's tier 1 target MSRV from 1.61 to 1.62. This change enables support for a broader range of `windows-core` versions, including the latest releases. This update is particularly beneficial to downstream libraries heavily utilizing `windows-bindgen`, as it reduces duplicate dependencies in their lock files.

The update is also beneficial to us as maintainers of `iana-time-zone` since the updated bindings are smaller and we will have an easier time tracking upstream.

@djc, from Chrono’s standpoint, is it acceptable to include this MSRV increase in another 0.1.x release? Alternatively, would you prefer we mark this clearly by releasing it as version 0.2.0?

I’m open to feedback and eager to find a path forward that aligns with everyone’s needs.

complexspaces:
Hey everyone 👋. I wanted to chime in as a downstream user of `iana-time-zone` (via `chrono`).

At work this is currently blocking us from cleanly migrating to the latest versions of `windows` and `windows-core` which we need to support new COM functionality because `iana-time-zone` is the one of the only third-party crates in our tree using something that isn't `windows-sys` or `windows-bindgen`. We would love to not introduce this duplication since it hurts compile times.

On my side as an OSS developer, I've worked around `windows-*` versioning conflicts in a few ways:
- Allowing a wider range of `windows-sys` versions and using trait glue to handle source incompatibility ([ref](https://github.com/1Password/arboard/pull/176)).
    - At a glance, I don't know if this approach would work for `iana-time-zone` today because it uses WinRT APIs as opposed to Win32 C ones.
- Use `windows-bindgen` with the `minimal` config option, which tells `windows-bindgen` to "vendor" types that would usually be provided via a `windows-core` import ([ref](https://github.com/1Password/sys-locale/pull/26))

If the MSRV arrangements here end up being problematic, I'd like to propose a different approach: Dropping the use of `Windows.Globalization.Calendar` and instead using [C APIs](https://learn.microsoft.com/en-us/windows/win32/api/timezoneapi/nf-timezoneapi-getdynamictimezoneinformation) that have been available since Windows 10 1703 (2016) per [this blog](https://devblogs.microsoft.com/oldnewthing/20210527-00/?p=105255) on converting time zone formats on Windows.  _This would require_ supporting a 1 year newer version of Windows 10 as opposed to the initial release, but in exchange I believe it would let `iana-time-zone` drop its `windows-core` dependency in favor of using purely `windows-bindgen`-generated code instead, unblocking all of this.

djc:
Bumping MSRV to 1.62 is fine at this point. Though, following through on @complexspaces points would be nice. (chrono itself already uses windows-bindgen too.)

astraw:
Thanks for the feedback, everyone. As I rarely use Windows, and even more rarely program low-level Windows-specific stuff, I really depend on your input.

These 1k+ lines of autogenerated code, with lots of `unsafe`, "just" to get the timezone on Windows seem suboptimal. What's not clear to me is if any of the ideas here, either @MarijnS95's updated PRs (thanks!) or @complexspaces idea of using the `timezoneapi.h` C API and `windows-bindgen`, would get us to a much more minimal code base (even in the more distant future in case it cannot be done right now). @MarijnS95, what do you think of the C API suggestion?

From my side, bumping MSRV to 1.62 would be fine with me, too.

lopopolo:
Merging this—thanks everyone for the thoughtful discussion and collaboration here!

I’ve opened #166 to continue the conversation around simplifying the Windows implementation and potentially dropping the `windows-core` dependency. Follow-ups and architectural ideas welcome there.

@MarijnS95 once this is in, I believe we can rebase #133 and land that pretty quickly. @astraw, I think we owe folks a release once all of these PRs are in.

lopopolo (approved):

PR #133, opened 2024-04-12
Bump `windows-core` to `0.56-0.61` range and regenerate with `0.59`

### LINKED ISSUE #158: Upgrade windows-core dependency
We are on v0.52.0. Upstream is on [v0.61.0](https://docs.rs/windows-core/0.61.0/windows_core/).

MarijnS95:
Closes #158

`windows-bindgen 0.59` now generates way less code since it no longer emits functions for which types are not enabled - that we don't use anyway!  And the added bonus is that this generated code is compatible with `windows-core` all the way back to release `0.56`, meaning we don't force everyone to immediately upgrade once they're already on `0.56`.


MarijnS95:
The required MSRV here is 1.62: https://docs.rs/crate/windows-core/0.56.0/source/Cargo.toml

lopopolo:
> Depends on #131

can you edit the base branch of this PR so that it is "stacked" on top of #131? It would help make the incremental delta easier to review.

MarijnS95:
> can you edit the base branch of this PR so that it is "stacked" on top of #131? It would help make the incremental delta easier to review.

That is unfortunately (but logically) only possible for branches directly in this repository. Since the _source_ branch of #131 lives on my fork instead, I cannot target it from a PR here (unless that PR would have been inside my fork, but then no-one would see it and we won't be able to update the base branch to `main` here once #131 is merged).

MarijnS95:
Notice that I've also tried generating bindings with `windows-bindgen 0.61`: https://github.com/MarijnS95/iana-time-zone/compare/windows-0.56..windows-bindgen-0.61

These are _mostly compatible_ except that they require the [`windows-collections` crate](https://crates.io/crates/windows-collections) which is currently only available for `windows-core 0.60` (`0.1.0`) and `windows-core 0.61` (`0.2.0`), unnecessarily bumping our compatible range.

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Changes
 - Bump MSRV (minimum supported rust version) to 1.62 ([#131](https://github.com/strawlab/iana-time-zone/pull/131))
-- Bump `windows-core` to `0.53-0.54` range ([#131](https://github.com/strawlab/iana-time-zone/pull/131))
+- Bump `windows-core` to `0.56-0.61` range ([#133](https://github.com/strawlab/iana-time-zone/pull/133))
```
lopopolo, line 10:
```suggestion
- Bump `windows-core` to `0.56-0.61` range (([#131](https://github.com/strawlab/iana-time-zone/pull/131)), [#133](https://github.com/strawlab/iana-time-zone/pull/133))
```

lopopolo (approved):

lopopolo:
MSRV build is failing due to:
```
   Compiling windows-result v0.1.1
error[E0635]: unknown feature `proc_macro_span_shrink`
  --> C:\Users\runneradmin\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\proc-macro2-1.0.52\src\lib.rs:92:30
   |
92 |     feature(proc_macro_span, proc_macro_span_shrink)
```

MarijnS95:
nightly*?

lopopolo:
oh, I don't care to support nightly with minimal versions. can you exclude this combo from the build matrix?

.github/workflows/rust.yml
```diff
@@ -20,17 +20,20 @@ jobs:
       matrix:
         runs-on: [ubuntu-22.04, windows-2022, macos-13]
         toolchain:
-          - "1.62"
+          - 1.62
```
lopopolo, line 23: This should be quoted so yaml doesn't parse it as a float

.github/workflows/rust.yml
```diff
@@ -20,17 +20,20 @@ jobs:
...
           - nightly
         versions:
           - ""
-          - "-Zminimal-versions"
+          - -Zminimal-versions
         exclude:
           # package `regex v1.11.1` cannot be built because it requires rustc
           # 1.65 or newer, while the currently active rustc version is 1.62.0
-          - toolchain: "1.62"
+          - toolchain: 1.62
```
lopopolo, line 32: Ditto

lopopolo (changes requested):

.github/workflows/rust.yml
MarijnS95: It's treated as a string by the parser either way?

astraw:
@MarijnS95 why did you close this? It seems a perfectly valid, and at this point probably preferred, solution to the Windows situation given the discovery you made and documented in #167.

I wonder if we could do something like the test in https://github.com/chronotope/chrono/blob/07216ae8fdc29073969920cb4957793dd47565e4/tests/win_bindings.rs
to validate that the autogenerated code is indeed autogenerated without requiring autogeneration for each build. I think this would also help with maintenance in the future and would reduce my concerns about the number of lines of code and unsafe usages.

astraw:
> @astraw because I sensed some very strong preference towards a "newer" and more confusing Windows API that requires (more) localization complexity _and_ doesn't run on some older Windows versions, all just to get rid of `windows-core`. This PR isn't moving towards that goal.

Last week it wasn't clear to me that I would find the "newer" API to be worse in practice. To me, at least, your work has shown that pretty clearly. Now I am in favor of merging this. @lopopolo @Kijewski any objections?

> The CI is already validating that the autogenerated code is actually generated, I added that when I introduced `windows-bindgen` support: https://github.com/strawlab/iana-time-zone/actions/runs/14131477375/job/39592670474?pr=133. No smoke & mirrors.

Ahh, great, I had missed that CI test, apologies.
 
> What do you mean with `autogeneration for each build`? Autogenerating is cheap at only 12 seconds of CI time, or do you consider that too much? To clear up any possible misunderstanding: this is _not a `build.rs`_ but only a separate tool to run when the `windows-bindgen` dependency or `bindings.txt` config change; the result is checked in and downstream (crate/repository) consumers don't need to run this again.

By "each build" I meant on running windows-bindgen on each downstream consumer's build. It is of course perfectly fine to run a 12 second check on each CI build. So this concern is perfect addressed, thank you again.

astraw:
Thanks everyone, especially @MarijnS95 . I released this now in 0.1.63.

complexspaces:
Thank you everyone for getting the new release out!

> Last week it wasn't clear to me that I would find the "newer" API to be worse in practice. To me, at least, your work has shown that pretty clearly.

I don't think the API is worse, its just a few more Win32 calls compared to the WinRT API that handles everything underneath for you. But, at the end of the day I don't have a strong opinion here and I'm equally happy to see `iana-time-zone` track newer versions of `windows-core`.

PR #139, opened 2024-09-16
Let dependabot monitor ci dependencies in ci actions

Kijewski:


astraw (approved):

astraw:
Nice, I didn't know dependabot could do this.

Issue #118, opened 2023-09-07
Setting the time zone via the TZ variable has no effect


teythoon:
On POSIX systems, the timezone can be overridden on a per-process basis by setting the `TZ` environment variable:

- https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_03
- https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html

Unfortunately, this has no effect on this crate.

astraw:
What platform are you on? How do you test this? On my PC:

```
andrew@pc iana-time-zone % cargo build --example get_timezone        
   Compiling iana-time-zone v0.1.57 (/Users/andrew/Documents/src/iana-time-zone)
    Finished dev [unoptimized + debuginfo] target(s) in 3.16s
andrew@pc iana-time-zone % target/debug/examples/get_timezone  
Europe/Berlin
andrew@pc iana-time-zone % TZ=EDT target/debug/examples/get_timezone 
America/New_York
```

teythoon:
I'm on Debian/Linux, mostly Bookworm (i.e. current stable):

```
% target/debug/examples/get_timezone
Europe/Berlin
% TZ=EDT target/debug/examples/get_timezone
Europe/Berlin
% TZ=Africa/Nairobi target/debug/examples/get_timezone
Europe/Berlin
% TZ=:Africa/Nairobi target/debug/examples/get_timezone
Europe/Berlin
```

astraw:
This crate developed organically and had no unifying concept behind handling the `TZ` environment variable. So thanks for bringing this up. I have now dug into this a bit.

According to the POSIX specification, there are two options for the `TZ` environment variable. First is `:characters`, which is handled in an "implementation-defined" manner. Second is a time zone rule literal, which defines the (bidirectional) mapping between Universal Time and the user's local time. We cannot convert a rule literal into an IANA time zone, so it seems there's nothing we can do there and I don't consider it further here. So, let's focus on the "implementation-defined" situation like `TZ=:Africa/Nairobi`.

It seems glibc, musl, Android, and macOS (at least) all follow a similar convention in which the local IANA time zone can be specified in the `TZ` environment variable:
 - glibc: https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html
 - musl: https://musl.libc.org/doc/1.1.24/manual.html
[...]

PR #144, opened 2024-11-24
on linux if TZ env var set and starts with ':', use as timezone

astraw:
This is an idea to address #118 on linux. What do folks think? @lopopolo @Kijewski

### DIFF
--- a/src/tz_linux.rs
+++ b/src/tz_linux.rs
@@ -1,6 +1,17 @@
 use std::fs::{read_link, read_to_string};
 
 pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
+    if let Some(tz) = std::env::var_os("TZ") {
+        match tz.into_string() {
+            Ok(mut tz) => {
+                if tz.starts_with(":") {
+                    return Ok(tz.split_off(1));
+                }
+            }
+            Err(_) => { /* not UTF-8, ignore TZ env var */ }
+        }
+    }
+
     etc_localtime()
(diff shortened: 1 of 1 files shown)


src/tz_linux.rs
```diff
@@ -1,6 +1,17 @@
 use std::fs::{read_link, read_to_string};
 
 pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
+    if let Some(tz) = std::env::var_os("TZ") {
```
lopopolo, line 4: Given that `env::var_os` is due to be made unsafe in Rust 2024, I'm not comfortable with this. If we do support reading from the TX env var, I'd rather it be gated by a feature and exposed via a new entry point.

At least in Artichoke, I'd not use this as is since my application must run its own hooks on env accesses and e.g. allow disabling all env accesses via a feature for sandboxing.

Kijewski: I'd make the same argument. Interacting with environment variables is IMHO out-of-scope of this library.

astraw: `env::var_os` remains safe in edition 2024. (It is `set_var` and `remove_var` that are [now unsafe](https://doc.rust-lang.org/edition-guide/rust-2024/newly-unsafe-functions.html).)

astraw:
Thanks for the feedback. I will close this.

I think a new entry point would probably be best done outside this crate, as I cannot see what the argument would be to have it here.

It seems you would recommend closing #118 with "Won't Fix"?

PR #146, opened 2025-01-27
Add support for tvOS, watchOS and visionOS

### LINKED ISSUE #71: Add support for tvOS (`aarch64-apple-tvos`) to Apple bindings
It looks like `core-foundation-sys` supports tvOS. I think this should be an easy fix apart from adding the `cargo check` to CI.

madsmtm:
By using `target_vendor = "apple"` instead of `any(target_os = "macos", target_os = "ios")` - the implementation uses CoreFoundation functions which are available on all Apple platforms.

Fixes #71.


src/lib.rs
```diff
@@ -55,7 +55,7 @@ mod ffi_utils;
 #[cfg_attr(target_os = "aix", path = "tz_aix.rs")]
 #[cfg_attr(target_os = "android", path = "tz_android.rs")]
 #[cfg_attr(target_os = "haiku", path = "tz_haiku.rs")]
-mod platform;
+mod tz_darwin;
```
lopopolo, line 58: This looks like a bad editor refactor. This module must always be named platform. Its contents are determined by the `#[path]` attribute in the chain of cfg above.

src/lib.rs
```diff
@@ -100,7 +100,7 @@ impl From<std::io::Error> for GetTimezoneError {
 /// about this function.
 #[inline]
 pub fn get_timezone() -> Result<String, GetTimezoneError> {
-    platform::get_timezone_inner()
+    tz_darwin::get_timezone_inner()
```
lopopolo, line 103: Please revert this change

lopopolo (changes requested):

src/lib.rs
madsmtm: Oops, indeed, `rust-analyzer` must've done it behind my back

Kijewski (approved):
I like it! Much better than trying to manually keep up with list of Apple OSes.

(I should really clean up the CI checks, though.)

madsmtm:
Btw, I chose the name `tz_darwin` because that's the name of the underlying OS, see also [`std::os::darwin`](https://doc.rust-lang.org/std/os/darwin/index.html).

lopopolo (approved):
thanks for fixing this up and expanding our platform support!

lopopolo:
Pulling in https://github.com/strawlab/iana-time-zone/pull/148 once it lands should get most of the builds passing.

lopopolo:
@madsmtm if you can rebase and resolve conflicts we can get this merged

madsmtm:
Hmm, I've rebased, but the CI didn't seem to run?

lopopolo:
@Kijewski @astraw can one of you cut a release?

PR #148, opened 2025-02-02
Fix up clippy errors from nightly on macos and linux targets

lopopolo:
no code changes in the unsafe macos bits, just shuffling some comments around to make the linter happy. I've included the additional safety context from https://github.com/strawlab/iana-time-zone/pull/147#issuecomment-2618070786.

This should fix the failing builds on main.

### DIFF
--- a/src/tz_linux.rs
+++ b/src/tz_linux.rs
@@ -89,7 +89,7 @@ mod openwrt {
             }
         }
 
-        timezone.ok_or_else(|| crate::GetTimezoneError::OsError)
+        timezone.ok_or(crate::GetTimezoneError::OsError)
     }
 
     #[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


src/tz_linux.rs
```diff
@@ -89,7 +89,7 @@ mod openwrt {
             }
         }
 
-        timezone.ok_or_else(|| crate::GetTimezoneError::OsError)
+        timezone.ok_or(crate::GetTimezoneError::OsError)
```
lopopolo, line 92: `crate::GetTimezoneError::OsError` is `Copy` which is why clippy says the closure is not needed.

src/tz_macos.rs
```diff
@@ -119,8 +122,8 @@ mod string_ref {
                 length,
             };
 
+            // SAFETY: `StringRef` is only ever created with a valid `CFStringRef`.
```
lopopolo, line 125: `clippy::undocumented_unsafe_blocks` wants the SAFETY comment on the block itself, not inside of it.

Kijewski (approved):
Thank you!

PR #150, opened 2025-03-19
Implement for OpenHarmony

ldm0:
OpenHarmony(or say [HarmonyOS NEXT](https://en.wikipedia.org/wiki/HarmonyOS_NEXT)) is forked from musl/Linux, but it doesn't have `/etc/localtime`, we have to use it's ["Time Service" function](https://gitee.com/openharmony/docs/blob/43726785b4033887cd1a838aaaca5e255897a71e/en/application-dev/reference/apis-basic-services-kit/time__service_8h.md) to get the timezone information.

Here is the test output:
```shell
$ cargo dinghy -d ohos test -Zbuild-std -- tests::get_current --nocapture
   Targeting platform auto-ohos-arm64-v8a and device 127.0.0.1:5555
   Compiling iana-time-zone v0.1.61 (/Users/donoughliu/code/iana-time-zone)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 1.49s
     Running unittests src/lib.rs (target/aarch64-unknown-linux-ohos/debug/deps/iana_time_zone-c96255da1500e341)
   Targeting platform auto-ohos-arm64-v8a and device 127.0.0.1:5555
  Installing iana_time_zone-c96255da1500e341 to 127.0.0.1:5555
     Running iana_time_zone-c96255da1500e341 on 127.0.0.1:5555

running 1 test
current: Asia/Shanghai
test tests::get_current ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.00s
```

### DIFF
--- a/src/tz_ohos.rs
+++ b/src/tz_ohos.rs
@@ -1,0 +1,47 @@
+//! OpenHarmony doesn't have `/etc/localtime`, we have to use it's "Time Service" to get the timezone information:
+//!
+//! - [API Refence](https://gitee.com/openharmony/docs/blob/43726785b4033887cd1a838aaaca5e255897a71e/en/application-dev/reference/apis-basic-services-kit/_time_service.md#oh_timeservice_gettimezone)
+
+use crate::ffi_utils::buffer::{tzname_buf, MAX_LEN};
+use crate::GetTimezoneError;
+use std::ffi::{c_char, CStr};
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+#[repr(C)]
+#[allow(non_camel_case_types)]
+#[allow(dead_code)]
+enum TimeService_ErrCode {
+    TIMESERVICE_ERR_OK = 0,
+    TIMESERVICE_ERR_INTERNAL_ERROR = 13000001,
+    TIMESERVICE_ERR_INVALID_PARAMETER = 13000002,
+}
+
+#[link(name = "time_service_ndk", kind = "dylib")]
+extern "C" {
+    fn OH_TimeService_GetTimeZone(timeZone: *mut c_char, len: u32) -> TimeService_ErrCode;
+}
+
+/// TODO: Change this `CStr::from_bytes_until_nul` once MSRV was bumped above 1.69.0
+fn from_bytes_until_nul(bytes: &[u8]) -> Option<&CStr> {
+    let nul_pos = bytes.iter().position(|&b| b == 0)?;
+    // SAFETY:
+    // 1. nul_pos + 1 <= bytes.len()
+    // 2. We know there is a nul byte at nul_pos, so this slice (ending at the nul byte) is a well-formed C string.
+    Some(unsafe { CStr::from_bytes_with_nul_unchecked(&bytes[..=nul_pos]) })
+}
+
+pub(crate) fn get_timezone_inner() -> Result<String, GetTimezoneError> {
+    let mut time_zone = tzname_buf();
+    // SAFETY:
+    // `time_zone` is a valid buffer with a length of 40 bytes.
+    let ret = unsafe {
+        OH_TimeService_GetTimeZone(time_zone.as_mut_ptr().cast::<c_char>(), MAX_LEN as u32 - 1)
+    };
+    if ret != TimeService_ErrCode::TIMESERVICE_ERR_OK {
+        return Err(GetTimezoneError::OsError);
+    }
+    from_bytes_until_nul(&time_zone)
+        .and_then(|x| x.to_str().ok())
+        .map(|x| x.to_owned())
+        .ok_or(GetTimezoneError::OsError)
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - bump msrv to 1.61 ([#157](https://github.com/strawlab/iana-time-zone/pull/157))
 
 ### Added
-- Added support for tvOS, watchOS and visionOS.
+- Added support for tvOS, watchOS and visionOS ([#146](https://github.com/strawlab/iana-time-zone/pull/146)).
+- implement OpenHarmony support ([#150](https://github.com/strawlab/iana-time-zone/pull/150))
 
 ## [0.1.61] - 2024-09-16
 
--- a/src/ffi_utils.rs
+++ b/src/ffi_utils.rs
@@ -4,7 +4,7 @@
 use std::ffi::CStr;
 
 /// A buffer to store the timezone name when calling the C API.
-#[cfg(any(test, target_vendor = "apple"))]
+#[cfg(any(test, target_vendor = "apple", target_env = "ohos"))]
 pub(crate) mod buffer {
     /// The longest name in the IANA time zone database is 32 ASCII characters long.
     pub const MAX_LEN: usize = 64;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,7 +34,11 @@
 #[allow(dead_code)]
 mod ffi_utils;
 
-#[cfg_attr(any(target_os = "linux", target_os = "hurd"), path = "tz_linux.rs")]
+#[cfg_attr(
+    any(all(target_os = "linux", not(target_env = "ohos")), target_os = "hurd"),
+    path = "tz_linux.rs"
+)]
+#[cfg_attr(all(target_os = "linux", target_env = "ohos"), path = "tz_ohos.rs")]
 #[cfg_attr(target_os = "windows", path = "tz_windows.rs")]
 #[cfg_attr(target_vendor = "apple", path = "tz_darwin.rs")]
 #[cfg_attr(


lopopolo:
@ldm0 can you point us in the direction of docs for ohos?

src/tz_ohos.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#[link(name = "time_service_ndk", kind = "dylib")]
+extern "C" {
+    fn OH_TimeService_GetTimeZone(timeZone: *mut std::ffi::c_char, len: u32)
+        -> TimeService_ErrCode;
+}
+
+pub(crate) fn get_timezone_inner() -> Result<String, GetTimezoneError> {
+    let mut time_zone = [0; 40];
+    let ret = unsafe { OH_TimeService_GetTimeZone(time_zone.as_mut_ptr(), time_zone.len() as u32) };
```
astraw, line 25: Please add a `// SAFETY:` comment for the `unsafe` as described [here](https://rust-lang.github.io/rust-clippy/stable/index.html#undocumented_unsafe_blocks).

Also, the docs are not perfectly clear to me: the `len` parameter is "without maximum limit", so should it actually be `time_zon.len()-1`? This is also the understanding I get when I read the error description for `TIMESERVICE_ERR_INVALID_PARAMETER `: "The time zone is null or the length of the time zone name (excluding the end character \0) is greater than or equal to len." It seems like including the `-1` will make the code safer in that it specifies the `time_zone` array will end with `\0`.

astraw (changes requested):
Thanks.

Can you please check that this passes clippy?

src/tz_ohos.rs
ldm0: Yes, have to admit the `-1` is needed from the doc. But the [official test](https://github.com/openharmony/time_time_service/blob/93ada192086efb7caf4033e6896e8fff3cd22943/test/unittest/native_test/time_service_native_test.cpp#L37) does use `time_zone.len()` directly, which is misleading. Anyway, using `len - 1` makes it safer.

src/tz_ohos.rs
```diff
@@ -0,0 +1,48 @@
...
+    TIMESERVICE_ERR_INVALID_PARAMETER = 13000002,
+}
+
+#[link(name = "time_service_ndk", kind = "dylib")]
+extern "C" {
+    fn OH_TimeService_GetTimeZone(timeZone: *mut std::ffi::c_char, len: u32)
+        -> TimeService_ErrCode;
+}
+
+/// TODO: Change this `CStr::from_bytes_until_nul` once MSRV was bumped above 1.72.0
```
lopopolo, line 24:
```suggestion
/// TODO: Change this `CStr::from_bytes_until_nul` once MSRV was bumped above 1.69.0
```

per the docs: https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_until_nul

src/tz_ohos.rs
```diff
@@ -0,0 +1,48 @@
...
+fn from_bytes_until_nul(bytes: &[u8]) -> Option<&CStr> {
+    let nul_pos = bytes.iter().position(|&b| b == 0)?;
+    // SAFETY:
+    // 1. nul_pos + 1 <= bytes.len()
+    // 2. We know there is a nul byte at nul_pos, so this slice (ending at the nul byte) is a well-formed C string.
+    Some(unsafe { CStr::from_bytes_with_nul_unchecked(&bytes[0..=nul_pos]) })
+}
+
+pub(crate) fn get_timezone_inner() -> Result<String, GetTimezoneError> {
+    let mut time_zone = [0; 41];
```
lopopolo, line 34:
```suggestion
    let mut time_zone: [c_char; 41] = [0; 41];
```

be explicit that this is a `c_char` to avoid problems with type inference and this alias'd type.

src/tz_ohos.rs
ldm0: Ah it's looks like [from_bytes_with_nul](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_with_nul)

src/tz_ohos.rs
```diff
@@ -0,0 +1,48 @@
...
+    let ret =
+        unsafe { OH_TimeService_GetTimeZone(time_zone.as_mut_ptr(), time_zone.len() as u32 - 1) };
+    if ret != TimeService_ErrCode::TIMESERVICE_ERR_OK {
+        return Err(GetTimezoneError::OsError);
+    }
+    let Some(tz_string) =
+        from_bytes_until_nul(&time_zone).map(|x| x.to_string_lossy().into_owned())
+    else {
+        return Err(GetTimezoneError::OsError);
+    };
```
lopopolo, line 45: [let-else is stabilized in Rust 1.65.0](https://doc.rust-lang.org/rust-by-example/flow_control/let_else.html) which is higher than this crate's MSRV.

```suggestion
    let tz_string = from_bytes_until_nul(&time_zone).ok_or(GetTimezoneError::OsError)?;
    let tz_string = tz_string.to_str().map_err(|_|  GetTimezoneError::OsError)?.to_owned();
```

src/tz_ohos.rs
ldm0: Done

src/tz_ohos.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn OH_TimeService_GetTimeZone(timeZone: *mut c_char, len: u32) -> TimeService_ErrCode;
+}
+
+/// TODO: Change this `CStr::from_bytes_until_nul` once MSRV was bumped above 1.69.0
+fn from_bytes_until_nul(bytes: &[u8]) -> Option<&CStr> {
+    let nul_pos = bytes.iter().position(|&b| b == 0)?;
+    // SAFETY:
+    // 1. nul_pos + 1 <= bytes.len()
+    // 2. We know there is a nul byte at nul_pos, so this slice (ending at the nul byte) is a well-formed C string.
+    Some(unsafe { CStr::from_bytes_with_nul_unchecked(&bytes[0..=nul_pos]) })
```
lopopolo, line 29:
```suggestion
    Some(unsafe { CStr::from_bytes_with_nul_unchecked(&bytes[..=nul_pos]) })
```

lopopolo (approved):

lopopolo:
@astraw with this and the apple vendor changes on trunk, we're probably due for a release

lopopolo:
@ldm0 can you rebase on top of https://github.com/strawlab/iana-time-zone/pull/154 and use the new buffer helpers in your code?

src/tz_ohos.rs
```diff
@@ -0,0 +1,45 @@
...
+    // 2. We know there is a nul byte at nul_pos, so this slice (ending at the nul byte) is a well-formed C string.
+    Some(unsafe { CStr::from_bytes_with_nul_unchecked(&bytes[..=nul_pos]) })
+}
+
+pub(crate) fn get_timezone_inner() -> Result<String, GetTimezoneError> {
+    let mut time_zone: [c_char; 41] = [0; 41];
+    // SAFETY:
+    // `time_zone` is a valid buffer with a length of 40 bytes.
+    let ret =
+        unsafe { OH_TimeService_GetTimeZone(time_zone.as_mut_ptr(), time_zone.len() as u32 - 1) };
```
lopopolo, line 37: using the new buffer helpers, you'd be doing `time_zone.as_mut_ptr().cast::<c_char>()` here.

src/tz_ohos.rs
```diff
@@ -0,0 +1,45 @@
...
+    if ret != TimeService_ErrCode::TIMESERVICE_ERR_OK {
+        return Err(GetTimezoneError::OsError);
+    }
+    from_bytes_until_nul(&time_zone)
```
lopopolo, line 41: ahh this is a subtle incompatibility! not likely to matter in practice but the buf is defined as c_char which isn't guaranteed to be the same as `u8`. Sorry for the thrash here. I remembered when I interact with FFI and C strings like this I typically declare the Rust-side storage as `u8` and then cast the pointer to `c_char` at the FFI callsite.

src/tz_ohos.rs
```diff
@@ -0,0 +1,45 @@
...
+fn from_bytes_until_nul(bytes: &[u8]) -> Option<&CStr> {
+    let nul_pos = bytes.iter().position(|&b| b == 0)?;
+    // SAFETY:
+    // 1. nul_pos + 1 <= bytes.len()
+    // 2. We know there is a nul byte at nul_pos, so this slice (ending at the nul byte) is a well-formed C string.
+    Some(unsafe { CStr::from_bytes_with_nul_unchecked(&bytes[..=nul_pos]) })
+}
+
+pub(crate) fn get_timezone_inner() -> Result<String, GetTimezoneError> {
+    let mut time_zone: [c_char; 41] = [0; 41];
```
lopopolo, line 33:
```suggestion
    let mut time_zone = [0u8; 41];
```

ldm0:
@lopopolo Done, might need to be rebased again after all CI passes though.

astraw (approved):

PR #151, opened 2025-03-22
make doctest no longer ignored

astraw:
This removes `ignored` from a doctest. The reason it was ignored was to not require the `chrono-tz` crate, but this can be done with a dev-dependency.


Kijewski (approved):
Good idea!

lopopolo (approved):

PR #152, opened 2025-03-22
Disable Android FFI helpers on non-Android targets

lopopolo:
Keep the helpers enabled in `test` on all platforms to make it easier to run these tests, which don't depend on the Android target itself.


Kijewski (approved):
:+1:

PR #153, opened 2025-03-23
Fix Apple timezone caching issue by resetting system timezone cache

### LINKED ISSUE #145: Changes to the system time zone are not visible in macOS

lopopolo:
Previously, changing the macOS system timezone while a process was running would not be reflected in subsequent calls to `get_timezone()`. This was due to `CFTimeZoneCopySystem` internally caching the timezone upon the first call.

This commit addresses the issue by explicitly calling `CFTimeZoneResetSystem` before each retrieval of the timezone, forcing CoreFoundation to invalidate and refresh its cached timezone value.

This approach incurs a minimal performance penalty (one lock and minor allocations) but ensures correctness and consistency with behavior observed on other platforms (Windows, Linux).

Closes #145.

cc @kathoum. This PR is the proposal to round out the investigation in https://github.com/strawlab/iana-time-zone/issues/145#issuecomment-2745934606.

### DIFF
--- a/examples/get_timezone_loop.rs
+++ b/examples/get_timezone_loop.rs
@@ -1,0 +1,13 @@
+use std::thread;
+use std::time::Duration;
+
+use iana_time_zone::{get_timezone, GetTimezoneError};
+
+const WAIT: Duration = Duration::from_secs(1);
+
+fn main() -> Result<(), GetTimezoneError> {
+    loop {
+        println!("{}", get_timezone()?);
+        thread::sleep(WAIT);
+    }
+}
--- a/src/tz_darwin.rs
+++ b/src/tz_darwin.rs
@@ -32,7 +32,9 @@ mod system_time_zone {
     //! create a safe wrapper around `CFTimeZoneRef`
 
     use core_foundation_sys::base::{CFRelease, CFTypeRef};
-    use core_foundation_sys::timezone::{CFTimeZoneCopySystem, CFTimeZoneGetName, CFTimeZoneRef};
+    use core_foundation_sys::timezone::{
+        CFTimeZoneCopySystem, CFTimeZoneGetName, CFTimeZoneRef, CFTimeZoneResetSystem,
+    };
 
     pub(crate) struct SystemTimeZone(CFTimeZoneRef);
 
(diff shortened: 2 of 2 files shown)


lopopolo:
I added a new `get_timezone_loop` example based on @kathoum's example in https://github.com/strawlab/iana-time-zone/issues/145#issue-2767308784 and ran it on my system while manually changing the timezone in system settings. 🎉 

Screenshot 2025-03-22 at 5 50 59 PM

astraw (approved):

astraw:
Nice work, thanks.

PR #156, opened 2025-03-23
Update Unicode license in cargo deny config

lopopolo:
Unicode 16.0.0 updated the license for data files from Unicode-2016-DFS to Unicode-3.0. Various crates across the ecosystem have been updating their SPDX specifiers to take this into account, including unicode-ident, which we pull in via syn and wasm-bindgen.


astraw (approved):

astraw:
Thanks again.

PR #157, opened 2025-03-23
Bump MSRV to 1.61.0, matching Chrono

lopopolo:
[Rust 1.61.0 came out in May 2022](https://blog.rust-lang.org/2022/05/19/Rust-1.61.0.html). CI is quite red, largely due to dep resolution, parsing, and compilation failures on the 1.48.0 toolchain.

Given our major reverse dep has increased MSRV, I feel we can do the same to ease the CI burden.

There are _a lot_ of CI jobs here and I haven't looked at all of them. @astraw, @Kijewski I could use your help determining if there is cruft here we can get rid of.

Additionally, on macOS, it looks like GitHub is moving in the direction of arm64 runners, which don't have a Rust toolchain in 1.48. Even in 1.61, the `aarch64-apple-darwin` target is only tier 2.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - main
   pull_request:
-    branches: [ '**' ]
+    branches: ["**"]
   schedule:
     # At 23:25 on Thursday.
     - cron: "25 23 * * 4"
@@ -24,6 +24,11 @@ jobs:
         versions:
           - ""
           - "-Zminimal-versions"
+        exclude:
+          # package `regex v1.11.1` cannot be built because it requires rustc
+          # 1.65 or newer, while the currently active rustc version is 1.61.0
+          - toolchain: "1.61"
+            versions: ""
     runs-on: ${{ matrix.runs-on }}
     steps:
       - name: Checkout
@@ -200,6 +205,11 @@ jobs:
         versions:
           - ""
           - "-Zminimal-versions"
+        exclude:
+          # package `cc v1.2.17` cannot be built because it requires rustc 1.63
+          # or newer, while the currently active rustc version is 1.61.0
+          - toolchain: "1.61"
+            versions: ""
     runs-on: ubuntu-latest
     env:
       RUSTFLAGS: -D warnings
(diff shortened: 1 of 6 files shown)


.github/workflows/rust.yml
```diff
@@ -18,26 +18,16 @@ jobs:
...
           - nightly
         versions:
           - ""
           - "-Zminimal-versions"
-        exclude:
-          # `windows-core >= 0.51` uses `edition = "2021"`, but users of the library can can use
-          # `windows-core == 0.50` to be rust 1.48 compatible.
-          - runs-on: windows-2022
-            toolchain: "1.48"
-            versions: ""
```
lopopolo, line 32: we haven't kept up to date with `windows-core` releases. Upstream started programatically tracking msrv here:

https://github.com/microsoft/windows-rs/blob/50009500b686f542530515aedba8334dc5b617b0/.github/workflows/msrv.yml#L37-L40

```yaml
      - name: Rust version
        run: rustup update --no-self-update 1.74 && rustup default 1.74
      - name: Check windows-core
        run:  cargo check -p windows-core --all-features
```

But this isn't historical and I'm not sure what range we should be using here. I filed https://github.com/strawlab/iana-time-zone/issues/158 for tracking this.

.github/workflows/rust.yml
```diff
@@ -18,26 +18,16 @@ jobs:
...
     runs-on: ${{ matrix.runs-on }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - uses: maxim-lobanov/setup-xcode@v1
-        if: matrix.runs-on == 'macos-13' && matrix.toolchain == '1.48'
-        with:
-          xcode-version: "13.4.1"
```
lopopolo, line 40: I think this can be nuked but I'm not sure what the context is.

lopopolo:
Audit build failure will be addressed by https://github.com/strawlab/iana-time-zone/pull/156 and the cross builds are sad, but everything else is green! This is ready for review.

.github/workflows/rust.yml
```diff
@@ -95,7 +90,7 @@ jobs:
       - name: Install Node
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: "lts/*"
```
lopopolo, line 93: for wasm builds, latest LTS should be fine and this removes a version string that needs to be kept up to date

astraw (approved):

astraw:
Yes, I concur that the cost of maintaining compatibility with 1.48 is not worth the benefit. (It could be done, but for what purpose?) Thanks for your work on this.

PR #159, opened 2025-03-23
haiku: improve C++ glue

lopopolo:
- Defend against NULL input buffer.
- Use auto and static_cast to clean up type conversions.
- Use BString::CopyInto API instead of raw memcpy.
- NUL terminate the buffer if possible.

I've bumped the crate version to prep for a release. Should also include https://github.com/strawlab/iana-time-zone/pull/154 before it gets cut.

I've also run this through clang-tidy and modernized things a bit.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,7 +1,7 @@
 ---
 BasedOnStyle: Google
 IndentWidth: 4
-ColumnLimit: 100
+ColumnLimit: 120
 ---
 Language: Cpp
 SortIncludes: false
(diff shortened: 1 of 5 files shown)


lopopolo:
@Kijewski would you be able to run the example in your haiku env?

lopopolo:
This PR will address https://github.com/strawlab/iana-time-zone/security/code-scanning/1

lopopolo:
@Kijewski ping on this (it's been a while, hope you're well!)

Noticed this at the bottom of my PR queue. Would love to get it in and cut a release!

PR #160, opened 2025-03-23
update core-foundation-sys requirement on macos

astraw:
This fixes an issue I missed in the review of #153 in that CFTimeZoneResetSystem was introduced in core-foundation-sys version 0.8.5. Since 0.8.5 is yanked, here we depend on 0.8.6.

This was silly to let this mistake get through considering the time we just spent making most of the CI tests pass.

@lopopolo can you double check this?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ fallback = []
 android_system_properties = "0.1.5"
 
 [target.'cfg(target_vendor = "apple")'.dependencies]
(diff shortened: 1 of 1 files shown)


lopopolo:
I did see some failures in CI related to the function not existing. I think if we had rebased the MSRV bump into the macOS PR before merging we would have caught this

lopopolo (approved):

PR #161, opened 2025-03-23
Upgrade to 2021 edition

lopopolo:
Since MSRV is now 1.61.0, we can compile with 2021 edition.

I followed the steps here:

- https://doc.rust-lang.org/edition-guide/editions/transitioning-an-existing-project-to-a-new-edition.html

I consider this low risk, but I don't think we compile all the platforms we support in CI.


Kijewski (approved):
:+1:

PR #163, opened 2025-03-23
Address high and medium severity zizmor findings

lopopolo:
Do our part to improve sofware supply chain security.

Zizmor is a security analysis tool for GitHub Actions workflows to ensure they are being used securely. An example finding is shown below:

```
warning[artipacked]: credential persistence through GitHub Actions artifacts
  --> ./.github/workflows/rust.yml:34:9
   |
34 |         - name: Checkout
   |  _________-
35 | |         uses: actions/checkout@v4
   | |_________________________________- does not set persist-credentials: false
   |
   = note: audit confidence → Low
```

I ran `zizmor .` on the codebase and made two changes:

- Drop GitHub token permissions by default. It looks like our existing jobs that require permissions already ask for them.
- Disable credential persistence when using `actions/checkout`. None of our jobs push anything back into the repository, so this aligns us to drop privs we don't need.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,6 +11,8 @@ on:
     - cron: "25 23 * * 4"
   workflow_dispatch:
 
+permissions: {}
+
 jobs:
   test:
     strategy:
@@ -33,6 +35,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          persist-credentials: false
       - name: Install Rust
         run: |
           rustup toolchain install ${{ matrix.toolchain }} --profile minimal --no-self-update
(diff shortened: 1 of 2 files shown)


Kijewski (approved):
That's nice!

Kijewski:
Thank you for finding & mentioning that tool! Maybe it would you useful to add the check to the CI itself?

```yaml
  Zizmor:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
        with:
          persist-credentials: false

      - uses: taiki-e/install-action@v2
        with:
          tool: zizmor@1

      - name: Run zizmor
        run: zizmor --format sarif . > results.sarif
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }} 

      - name: Upload SARIF file
        uses: github/codeql-action/upload-sarif@v3
        with:
          sarif_file: results.sarif
          category: zizmor
```

lopopolo:
> Thank you for finding & mentioning that tool! Maybe it would you useful to add the check to the CI itself?

Done, using the suggested integration from upstream: https://woodruffw.github.io/zizmor/usage/#use-in-github-actions.

PR #164, opened 2025-03-24
AIX: remove unused code

### LINKED ISSUE #162: Revisit AIX platform support
This function is unused but was intentionally added during #57 after code review feedback:

https://github.com/strawlab/iana-time-zone/blob/e6bc833bf1d7d2c93c3e6b27d7f69e4ce0edae49/src/tz_aix.rs#L9-L26

Use of `std::env::var` will panic if the TZ env var is not valid UTF-8. To guard against this, the code should use `env::var_os` and a fallible conversion of `OsString` to `String`.

https://github.com/strawlab/iana-time-zone/blob/e6bc833bf1d7d2c93c3e6b27d7f69e4ce0edae49/src/tz_aix.rs#L6

suggested patch:

```rust
    let timezone = env::var_os("TZ").ok_or(crate::GetTimezoneError::OsError)?;
    timezone.into_string().map_err(|_| crate::GetTimezoneError::OsError)
}
```

astraw:
This fixes #162

### DIFF
--- a/src/tz_aix.rs
+++ b/src/tz_aix.rs
@@ -5,22 +5,3 @@ use std::io::{BufRead, BufReader};
 pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
     env::var("TZ").map_err(|_| crate::GetTimezoneError::OsError)
 }
-
-fn read_environment() -> Result<String, crate::GetTimezoneError> {
-    // https://www.ibm.com/docs/en/aix/7.2?topic=files-environment-file
-
-    let file = OpenOptions::new().read(true).open("/etc/environment")?;
-    let mut reader = BufReader::new(file);
-    let mut line = String::with_capacity(80);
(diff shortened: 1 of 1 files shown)


src/tz_aix.rs
```diff
@@ -5,22 +5,3 @@ use std::io::{BufRead, BufReader};
 pub(crate) fn get_timezone_inner() -> Result<String, crate::GetTimezoneError> {
     env::var("TZ").map_err(|_| crate::GetTimezoneError::OsError)
```
lopopolo, line 6: while you're in here, can you also apply the patch to avoid panicking on invalid UTF-8 in the `OsString`?

lopopolo (approved):
LGTM!

src/tz_aix.rs
lopopolo: whoops, just read your other comment. I stand corrected!

PR #173, opened 2025-08-19
Add support for VxWorks

hax0kartik:
This adds support for VxWorks in iana-time-zone. The added functionality is basically just a wrapper around VXAL crate where the actual implementation lives. VXAL is an internal crate which is automatically added to the rustc invocation by our custom rustc wrapper. Hence, there is no need to add anything to the toml file.

Please let me know if there are any concerns.


lopopolo:
If that's the case, I don't think I'm willing to accept this PR. I am not comfortable shipping a crate to crates.io which depends on an unpublished commercial crate. In the short term, I recommend you maintain an internal fork.

I think there is probably space to design an api here that lets uncommon platforms inject a custom implementation just like the getrandom crate does.

Please open an issue and we can discuss design and whether the other maintainers are aligned to this plan.

hax0kartik:
Sure, thanks!

